PR #6, opened 2021-04-06
Efficient diffing for OrderedSet

numist:
### Description
Adds an overload of `difference(from:)` that implements a linear time algorithm (same as used by Foundation's `NSOrderedSet`, SwiftUI's `List`, and UIKit's Diffable Data Source), as well as an implementation of `applying(:)` that is linear, though slower than if `OrderedSet` conformed to `RangeReplaceableCollection`.

### Detailed Design
```swift
extension OrderedSet {
  /// Returns the collection difference between the parameter and the
  /// receiver, using an algorithm specialized to exploit fast membership
  /// testing and the member uniqueness guarantees of `OrderedSet`.
  ///
  /// - Complexity: O(`self.count + other.count`)
  @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
  public func difference(
    from other: Self
  ) -> CollectionDifference<Element>

  /// Applies the given difference to this collection.
  ///
  /// - Parameter difference: The difference to be applied.
  ///
  /// - Returns: An instance representing the state of the receiver with the
  ///   difference applied, or `nil` if the difference is incompatible with
  ///   the receiver's state.
  ///
  /// - Complexity: O(*n* + 2 × (*n* + *c*)), where *n* is `self.count` and *c*
  ///   is the number of changes contained by the parameter.
  @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
  public func applying(_ difference: CollectionDifference<Element>) -> Self?
}
```

### Documentation
As shown above:

* `OrderedSet.difference(from:)` has headerdoc that refines the documentation from `BidirectionalCollection.difference(from:)` (which it overloads)
* `OrderedSet.applying(:)` has headerdoc that duplicates `RangeReplaceable.applying(:)` but with a slower `Complexity` 😬

### Testing
Pretty exhaustive correctness and performance testing added in `OrderedSet Diffing Tests.swift`.

### Performance
All unit tests perform a diff with a box type with counters for calls to `==` and `hash(into:)`, and then validate that both are < nlogn. Despite advertising linear performance, checking "less than the next meaningful order of magnitude" allows the non-deterministic tests some wiggle room for constant factors, multiple linear passes, and bad hash luck while still proving performance far superior to traditional diffing algorithms, which run in O(n**2) worst case (represented in the unit tests).

The implementation of `applying(:)` is still O(n), but definitely exposes room for improvement in the stdlib—a single-pass implementation could exist for all collection types with an O(1) `append(:)` but implementing it here would duplicate [`RangeReplaceableCollection.applying(:)`](https://github.com/apple/swift/blob/main/stdlib/public/core/Diffing.swift) and all of its support functions.

### Source Impact
Solely additive changes.

### Checklist
- [X] I've read the [Contribution Guidelines](../CONTRIBUTING.md)
- [X] I've added tests covering all new code paths my change adds to the project (to the extent possible).
- [X] I've verified that my change does not break any existing tests or introduce unexpected benchmark regressions.
- [X] I've updated the documentation (if appropriate).

### DIFF
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
@@ -75,5 +75,27 @@ extension OrderedSet {
     }
     return CollectionDifference(changes)!
   }
+
+  /* Replicating RangeReplaceableCollection.applying here would involve
+   * duplicating a ton of IPI from the stdlib. Dropping down to Array and back
+   * to OrderedSet is still O(n)-ish which still beats the naïve application
+   * algorithm of "remove, then insert"
+   */
+  /// Applies the given difference to this collection.
+  ///
+  /// - Parameter difference: The difference to be applied.
+  ///
+  /// - Returns: An instance representing the state of the receiver with the
+  ///   difference applied, or `nil` if the difference is incompatible with
+  ///   the receiver's state.
+  ///
+  /// - Complexity: O(*n* + 2 × (*n* + *c*)), where *n* is `self.count` and *c*
+  ///   is the number of changes contained by the parameter.
+  @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+  public func applying(_ difference: CollectionDifference<Element>) -> Self? {
+    guard let array = Array(self).applying(difference) else { return nil }
+    let result = OrderedSet(array)
+    return result.count == array.count ? result : nil
+  }
 }
 
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
@@ -89,7 +89,7 @@ extension OrderedSet {
   ///   difference applied, or `nil` if the difference is incompatible with
   ///   the receiver's state.
   ///
-  /// - Complexity: O(*n* + 2 × (*n* + *c*)), where *n* is `self.count` and *c*
+  /// - Complexity: O(*n* + *c*), where *n* is `self.count` and *c*
   ///   is the number of changes contained by the parameter.
   @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
   public func applying(_ difference: CollectionDifference<Element>) -> Self? {
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
@@ -57,12 +57,12 @@ extension OrderedSet {
             assert(y == axinb)
             // `a[x]` == `b[y]`
             x += 1; y += 1
-          } else if byina - x >= axinb - y {
-            // `b[y]` exists further away from the current position in `a` than `a[x]` does in `b`
-            remove()
-          } else {
+          } else if byina - x < axinb - y {
             // `a[x]` exists further away from the current position in `b` than `b[y] does in `a`
             insert()
+          } else {
+            // `b[y]` exists further away from the current position in `a` than `a[x]` does in `b`
+            remove()
           }
         } else {
           // `b[y]` does not exist in `a`, the element must have been inserted
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
@@ -14,7 +14,7 @@ extension OrderedSet {
   /// receiver, using an algorithm specialized to exploit fast membership
   /// testing and the member uniqueness guarantees of `OrderedSet`.
   ///
-  /// - Complexity: O(`left.count + right.count`)
+  /// - Complexity: O(`self.count + other.count`)
   @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
   public func difference(
     from other: Self
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
@@ -1,0 +1,79 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Collections open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+extension OrderedSet {
+  /// Returns the collection difference between the parameter and the
+  /// receiver, using an algorithm specialized to exploit fast membership
+  /// testing and the member uniqueness guarantees of `OrderedSet`.
+  ///
+  /// - Complexity: O(`left.count + right.count`)
+  @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+  public func difference(
+    from other: Self
+  ) -> CollectionDifference<Element> {
+    /* While admitting that variables with names like "a", "b", "x", and "y" are not especially readable, their use (and meaning) is standard in the diffing literature and familiarity with that literature will help if you're reading this code anyway. */
+    let a = other // The collection we're diffing "from". An element present in this collection but not the other is a "remove"
+    var x = 0 // The working index into `a`
+    let b = self // The collection we're diffing "to". An element present in this collection but not the other is an "insert"
+    var y = 0 // The working index into `b`
+
+    var changes = Array<CollectionDifference<Element>.Change>()
+    func remove() {
+      let ax = a[x]
+      changes.append(.remove(offset: x, element: ax, associatedWith: b.lastIndex(of: ax)))
+      x += 1
+    }
+    func insert() {
+      let by = b[y]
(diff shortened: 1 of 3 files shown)


634750802:
The link https://github.com/apple/swift-collections/CONTRIBUTING.md seems not a valid url.

Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
```diff
@@ -0,0 +1,101 @@
...
+  ///
+  /// - Returns: An instance representing the state of the receiver with the
+  ///   difference applied, or `nil` if the difference is incompatible with
+  ///   the receiver's state.
+  ///
+  /// - Complexity: O(*n* + 2 × (*n* + *c*)), where *n* is `self.count` and *c*
+  ///   is the number of changes contained by the parameter.
+  @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+  public func applying(_ difference: CollectionDifference<Element>) -> Self? {
+    guard let array = Array(self).applying(difference) else { return nil }
```
lorentey, line 96: Using the `elements` "view" would get rid of the initial copy, making this pretty efficient in the happy case. (`OrderedSet.init(_:)` is already able to simply adopt array storage instead of copying out.)

```suggestion
    guard let array = self.elements.applying(difference) else { return nil }
```

lorentey, line 96: Using the `elements` "view" here would get rid of the initial copy, making this pretty efficient in the happy case. (`OrderedSet.init(_:)` is already able to simply adopt array storage instead of copying out elements one by one.)

```suggestion
    guard let array = self.elements.applying(difference) else { return nil }
```

Tests/OrderedCollectionsTests/OrderedSet/OrderedSet Diffing Tests.swift
```diff
@@ -0,0 +1,142 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import OrderedCollections
+import CollectionsTestSupport
+
+class MeasuringHashable: Hashable {
```
lorentey, line 16: 👍 

(D'oh, I forgot to port the MinimalHashable types from StdlibUnittest!)

Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
```diff
@@ -0,0 +1,101 @@
...
+//===----------------------------------------------------------------------===//
+
+extension OrderedSet {
+  /// Returns the collection difference between the parameter and the
+  /// receiver, using an algorithm specialized to exploit fast membership
+  /// testing and the member uniqueness guarantees of `OrderedSet`.
+  ///
+  /// - Complexity: O(`self.count + other.count`)
+  @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+  public func difference(
```
lorentey, line 19: I wonder how performance would change if we made these `@inlinable @inline(never)`. 🤔

Tests/OrderedCollectionsTests/OrderedSet/OrderedSet Diffing Tests.swift
```diff
@@ -0,0 +1,142 @@
...
+    MeasuringHashable.hashChecks = 0
+
+    let _ = OrderedSet(a.map({MeasuringHashable($0)})).difference(from: OrderedSet(b.map({MeasuringHashable($0)})))
+    let n = a.count + b.count
+
+    /* Expect linear performance, which we can fence in testing as
+     * "less than nlogn" since diffing generally tends to be at least n**2
+     */
+    expectLessThan(MeasuringHashable.equalityChecks, n * n.bitWidth)
+    expectLessThan(MeasuringHashable.hashChecks, n * n.bitWidth)
```
lorentey, line 49: 💯💯💯

lorentey (approved):
This is excellent, thank you! 💯 

I'll run the new benchmarks with/without the improved methods offline -- I'm curious how well the benchmarking package handles these. (Esp. about how much noise there will be due to random input.)

lorentey:
@swift-ci test

Tests/OrderedCollectionsTests/OrderedSet/OrderedSet Diffing Tests.swift
numist: Why doesn't the stdlib have a logarithm function anyway?

Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
numist: How does one run a single benchmark? The entire suite takes about 2000s on my machine 😬

Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
numist: Oh excellent

drexin:
@swift-ci test

numist:
> I'll run the new benchmarks with/without the improved methods offline -- I'm curious how well the benchmarking package handles these. (Esp. about how much noise there will be due to random input.)

In my experience randomized inputs to diffing algorithms introduce less jitter than one would expect—most algorithms tend to be optimized for inputs where `b` is clearly derived from `a` so truly random inputs just invoke their worst case performance (Myers) or edit script length (Heckel). Myers even does fairly well against random inputs when the alphabet is small ([measured ~34k comparisons when diffing two binary strings with n = 500](https://github.com/numist/Diffing-Explorations#appendix-data) compared to a worst case 250k) since strings derived from a smaller alphabet are more likely to contain matches.

The algorithm implemented here compromises edit script length in favour of performance; an example of this can be found in `test_fuzzerDiscovered01`, where diffing a → b produces a larger edit script than diffing b → a:

``` swift
  func test_fuzzerDiscovered01() {
     // b → a X  X  X  X  X        X
     // a → b X  X  X  X  X  X     X
     let a = [3, 7, 2, 5, 9, 1, 8, 4]
     let b = [4, 2, 1, 7, 8, 6]
     // a → b X  X  X  X     X
     // b → a X  X     X     X
     _validate(from: a, to: b, mutations: -1)
   }
```

Sources/OrderedCollections/OrderedSet/OrderedSet+Diffing.swift
lorentey: You can use the `--filter` and `--exclude` options to only run tasks that match (or don't match) a specific regex!

(Alternatively, you can set up a custom library JSON that configures the charts you want to see, then run it with `library run --library foo.json`. It might be overkill for a couple of charts though!)

Tests/OrderedCollectionsTests/OrderedSet/OrderedSet Diffing Tests.swift
lorentey: BinaryInteger does have a customization point for it, it just isn't public yet -- it's currently called `_binaryLogarithm()`. (By default it is defined in terms of `leadingZeroBitCount`.)

numist:
Unfortunately the benchmark tool seems to die somewhere between 64k and 128k when measuring `Collection.difference(from:)`:

```
% swift run -c release swift-collections-benchmark run results --cycles 5 --filter " diff "
[0/0] Build complete!
Running 2 tasks on 76 sizes from 1 to 1M:
  OrderedSet<Int> previous diff computation
  OrderedSet<Int> diff application
Output file: /Volumes/stuff/numist/Code/oss/Swift Collections/results
Appending to existing data (if any) for these tasks/sizes.

Collecting data:
  1.2.4...8...16...32...64...128...256...512...1k...2k...4k...8k...16k...32k...64k..fish: Job 1, 'swift run -c release swift-coll…' terminated by signal SIGKILL (Forced quit)
```

But that does give us enough data to see the Myers implementation complexity lifting off:

chart

numist:
@swift-ci test

lorentey:
It seems to me we can enable specializations in a subsequent refinement. (My secret plan is that having the benchmarks in place will be a nice test case for the `results compare` command of the benchmarking facility.)

It looks like you were right about jitter! It is higher than usual, but it's not overly huge. 

(I find it interesting how the sample mean seems to be close to the middle of the "error" bands -- in less random cases, it is usually hugging the minimum, rather than being centered between the min and the mean+2*stddev curve. If I knew anything about statistics, this would probably indicate something about how the measured distribution differs from the usual case. 😇)

numist:
> (I find it interesting how the sample mean seems to be close to the middle of the "error" bands -- in less random cases, it is usually hugging the minimum, rather than being centered between the min and the mean+2*stddev curve. If I knew anything about statistics, this would probably indicate something about how the measured distribution differs from the usual case. 😇)

If you want a nice rabbit hole, you could try drawing the graphs as a [Density Line Chart](https://arxiv.org/pdf/1808.06019.pdf)—the mean may not represent a plausible series itself anyway (if, say, the results are highly bimodal)!

lorentey:
I have had a todo item for a while to implement scatter plots -- the lines are meaningless anyway, and as long as we aren't rendering too many benchmarks, the points would probably show up great. (The last time I tried it was when I was still doing these with pivot tables in Excel. 🙈)

I've also been meaning to dust of some stats textbooks. I'm sure I'll get around to it eventually. 😇

PR #21, opened 2021-04-13
Add the @discardableResult attribute to OrderedSet.insert(_:at:)

### LINKED ISSUE #19: OrderedSet.insert should be @discardableResult
Often I just need to insert an element into the OrderedSet. For now I need to prefix that with '_ = ' which in my option should be avoided. Please consider adding @discardableResult to 

insert(_ item: Element, at index: Index) -> (inserted: Bool, index: Index)

Best, Felix

kielgillard:
Fixes #19.

It does not seem appropriate to write a test for this change. However, I am open to other points of view and will gladly own further work.

### Checklist

### DIFF
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+Insertions.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+Insertions.swift
@@ -152,6 +152,7 @@ extension OrderedSet {
   ///    type, if it implements high-quality hashing. (Insertions need to make
   ///    room in the storage array to add the inserted element.)
   @inlinable
+  @discardableResult
   public mutating func insert(
     _ item: Element,
     at index: Index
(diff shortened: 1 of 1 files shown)


lorentey (approved):
This looks good, thank you!

It would be useful to see if there any other APIs in the package that are currently missing this attribute -- if we missed one, we may have missed others.

lorentey:
@swift-ci test

kielgillard:
I did wonder if the following functions should have this attribute, too:
`updateOrAppend(_ item: Element) -> Element?`
`updateOrInsert(_ item: Element, at index: Index) -> (originalMember: Element?, index: Index)`

lorentey:
@kielgillard I think so, if only by analogue to `SetAlgebra`'s `update(with:)` requirement. Can you add those to this PR? 😊

(I wonder how often these would be used without looking at the return value, but there is nothing wrong with discarding it, either -- and I don't think we want to force people to do the `_ = foo.updateOrAppend(...)` dance to achieve that.)

kielgillard:
@swift-ci test

kielgillard:
@lorentey done.

My intuition is people will probably be using these "update or" APIs specifically for the return value but, like you said, there's no need for the API to be a jerk about it in case you don't want it!

lorentey:
Looks good, thanks! 👍 

@swift-ci test

PR #22, opened 2021-04-13
Add missing final keyword

DimaMishchenko:
- Added missing `final` keyword
- Fixed small typo in `MutableCollectionTests` class name

### Checklist

### DIFF
--- a/Benchmarks/Benchmarks/CppBenchmarks.swift
+++ b/Benchmarks/Benchmarks/CppBenchmarks.swift
@@ -12,7 +12,7 @@
 import CollectionsBenchmark
 import CppBenchmarks
 
-internal class CppVector {
+internal final class CppVector {
   var ptr: UnsafeMutableRawPointer?
 
   init(_ input: [Int]) {
@@ -33,7 +33,7 @@ internal class CppVector {
   }
 }
 
-internal class CppDeque {
+internal final class CppDeque {
   var ptr: UnsafeMutableRawPointer?
 
   init(_ input: [Int]) {
@@ -54,7 +54,7 @@ internal class CppDeque {
   }
 }
 
-internal class CppUnorderedSet {
+internal final class CppUnorderedSet {
   var ptr: UnsafeMutableRawPointer?
 
   init(_ input: [Int]) {
@@ -75,7 +75,7 @@ internal class CppUnorderedSet {
   }
 }
 
-internal class CppUnorderedMap {
+internal final class CppUnorderedMap {
   var ptr: UnsafeMutableRawPointer?
 
   init(_ input: [Int]) {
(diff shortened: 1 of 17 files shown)


lorentey:
👍 

I don't see anything wrong with adding explicit `final` modifiers, but to be honest I don't see why we'd need to add them either. Can you explain why are you suggesting this change? 

E.g., is it only to make things explicit (like the way this codebase tends to spell out `internal` even though it's the default), or is there a practical effect I should know about?

lorentey:
@swift-ci test

DimaMishchenko:
@lorentey yes, sure. 
- explicit declaration, as you mentioned
- small performance optimization by reducing dynamic dispatch

lorentey:
@DimaMishchenko do you have a way to quantify the performance optimization? E.g., does it show up in benchmarks? Have you seen a codegen improvement?

kylemacomber:
I think it makes sense to make public types final if they aren't intended to be subclassed. It's my understanding that Swift will "infer" the final when possible, for example on any non-public classes that don't have subclasses (at least if you build with WMO on).

DimaMishchenko:
@lorentey first of all, sorry for long feedback.

What @kylemacomber said should be correct (at least according to [this](https://github.com/apple/swift/blob/main/docs/OptimizationTips.rst#advice-if-wmo-is-enabled-use-internal-when-a-declaration-does-not-need-to-be-accessed-outside-of-module) and [this](https://developer.apple.com/swift/blog/?id=27)).

Which should mean that changes are only useful form code-style perspective. 

But I did some additional performance testing for `Deque` (because `_DequeBuffer` was changed to `final`) to make sure that all works as expected. 
Did comparison of `Deque` with `final` `_DequeBuffer` vs non-final.

First, simply tested `append` operation with performance test (using `measure`).
``` swift
func testAppend() {
  var deque = Deque<Int>()
    
  let options = XCTMeasureOptions()
  options.iterationCount = 100
    
  measure {
    (0...1000000).forEach { _ in deque.append(1) }
  }
}
```
Results:
- Without WMO.
  - final - 0.775
  - non-final - 0.778 sec
- With WMO (-O)
  - final - 0.024
  - non-final - 0.026 sec  

Then, benchmark test `append` (`max-size` - 16M, `cycles` - 20).
01 Deque: append


Difference is quite small (and can be just accuracy error). 
Now for me it seems that changes are only useful to keep same style and do things explicit. 

So @lorentey if you think that this PR is unnecessary, please feel free to close it 😀

jjatie:
On the side of adding the `final` annotation: [swift/OptimizationTips](https://github.com/apple/swift/blob/main/docs/OptimizationTips.rst#id17) recommends it. Whether or not it has a meaningful impact for performance here, according to the linked article it is seems to be good general advice.

lorentey:
The recommendation in OptimizationTips is mostly out of date. It was added in 2015, predating Swift 3's introduction of the [`open` qualifier](https://github.com/apple/swift-evolution/blob/master/proposals/0117-non-public-subclassable-by-default.md).

PR #26, opened 2021-04-16
[OrderedDictionary] Document encoding format

### LINKED ISSUE #12: Codable Support
I was trying to use the OrderedDictionary in a Codable struct.
I have some API requests that I receive a few dictionaries and I would like to preserve that order.
Previously I have a property like this.

`let foos: [String: Foo]`

And the parse works just fine, but the order of the items isn't guaranteed.

So I changed to this,

`let foos: OrderedDictionary<String, Foo>`

and I get this error.

```
"Expected to decode Array<Any> but found a dictionary instead."
```

It would be great if you give support Codable parsing.

lorentey:
Unlike the regular unordered `Dictionary`, `OrderedDictionary` always encodes/decodes itself as alternating key/value pairs in an unkeyed container. Document this fact (and the reason behind it) in the `Decodable` implementation.

Resolves #12.

### Checklist


lorentey:
@swift-ci test

PR #32, opened 2021-05-02
Replace Index with Int where possible

### LINKED ISSUE #30: OrderedSet append does not return the same type as insert
The API for insert in an `OrderedSet` is
```swift
public mutating func insert(_ item: Element, at index: Index) -> (inserted: Bool, index: Index)
```

and the one for append is
```swift
public mutating func append(_ item: Element) -> (inserted: Bool, index: Int)
```

I understand the return types are technically the same because `Index` is `Int`, always, but from a semantic point of view, shouldn’t the append method return a `(inserted: Bool, index: Index)` tuple?

### Information

- **Package version:** Swift Collection `0.0.2`
- **Platform version:** macOS `11.3`
- **Swift version:**
```text
Apple Swift version 5.4 (swiftlang-1205.0.26.9 clang-1205.0.19.55)
Target: x86_64-apple-darwin20.4.0
```

### Checklist

### Steps to Reproduce
N/A

### Expected behavior
`OrderedSet`’s `append` method to return `(inserted: Bool, index: Index)`.

### Actual behavior
It returns `(inserted: Bool, index: Int)`.

Frizlab:
Closes #30. I tried to find all instances of Int where Index should have been used in the Source. I did not search the tests because the issue is semantic only, and my changes only have impact on the public API.

### Checklist

### DIFF
--- a/Sources/DequeModule/Deque+MutableCollection.swift
+++ b/Sources/DequeModule/Deque+MutableCollection.swift
@@ -22,7 +22,7 @@ extension Deque: MutableCollection {
   /// - Complexity: O(1) when this instance has a unique reference to its
   ///    underlying storage; O(`count`) otherwise.
   @inlinable
-  public mutating func swapAt(_ i: Index, _ j: Index) {
+  public mutating func swapAt(_ i: Int, _ j: Int) {
     precondition(i >= 0 && i < count, "Index out of bounds")
     precondition(j >= 0 && j < count, "Index out of bounds")
     _storage.ensureUnique()
--- a/Sources/DequeModule/Deque+RandomAccessCollection.swift
+++ b/Sources/DequeModule/Deque+RandomAccessCollection.swift
@@ -28,7 +28,7 @@ extension Deque: RandomAccessCollection {
   /// - Complexity: O(1)
   @inlinable
   @inline(__always)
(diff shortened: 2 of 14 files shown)


xwu:
Mmm, I'm not certain that the benefits outweigh the drawbacks here with this change. `Index` is explicitly an associated type, and it's arguable that readers _want_ to know what the actual associated type is. Personally I prefer the status quo.

Frizlab:
@xwu In any case the public API is inconsistent now (`Index` and `Int` are both used; see #30 for an example). I do think one of both should be chosen ;-)

Reading the source code of Swift’s `Array`, it seems `Int` is used, not `Index`, so maybe using `Int` would indeed be better for consistency with Swift itself.

I personally slightly prefer `Index` over `Int`, but I’d happy to update this PR to use `Int` instead of `Index` if the maintainers think it’s best.

lorentey:
The use of `Index` is mostly a leftover from the initial implementations where these collections used custom index types. 

I'd prefer if we went the other way and used `Int` everywhere -- the fact that these have integer indices carry additional semantic meaning, and it makes sense not to hide this information behind a typealias.

Frizlab:
Done

lorentey:
@swift-ci test

PR #38, opened 2021-05-17
Fix compilation under BUILD_LIBRARY_FOR_DISTRIBUTION flag

srdanrasic:
When building this package with `BUILD_LIBRARY_FOR_DISTRIBUTION` set to `YES`, the compilation fails in number of places with errors listed further below. You can reproduce the issue by attempting the following from the package root directory:

```
xcodebuild -workspace . -scheme Collections BUILD_LIBRARY_FOR_DISTRIBUTION=YES
```

Suggestions in compiler errors do not really work, but after some experimenting and from looking into stdlib, I found out that adding `@frozen` attribute to the problematic types fixes the problem.

I'm not sure if this is the right solution, but I'm happy to update my PR if you could point me into the right direction. I don't actually need support for library evolution, but I do need to be able to compile this package into an xcframework.
 

```swift
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._UnsafeHandle.swift:31:12: error: 'let' property '_header' may not be initialized directly; use "self.init(...)" or "self = ..." instead
      self._header = header
           ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._UnsafeHandle.swift:16:9: note: '_header' declared here
    let _header: UnsafeMutablePointer<_DequeBufferHeader>
        ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._UnsafeHandle.swift:32:12: error: 'let' property '_elements' may not be initialized directly; use "self.init(...)" or "self = ..." instead
      self._elements = elements
           ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._UnsafeHandle.swift:18:9: note: '_elements' declared here
    let _elements: UnsafeMutablePointer<Element>
        ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._UnsafeHandle.swift:34:12: error: 'let' property '_isMutable' may not be initialized directly; use "self.init(...)" or "self = ..." instead
      self._isMutable = isMutable
           ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._UnsafeHandle.swift:21:9: note: '_isMutable' declared here
    let _isMutable: Bool
        ^

/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque+Sequence.swift:31:21: error: 'self' used before 'self.init' call or assignment to 'self'
      self._storage = _storage
                    ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque+Sequence.swift:32:22: error: 'self' used before 'self.init' call or assignment to 'self'
      self._nextSlot = start
                     ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque+Sequence.swift:33:21: error: 'self' used before 'self.init' call or assignment to 'self'
      self._endSlot = end
                    ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque+Sequence.swift:34:5: error: 'self.init' isn't called on all paths before returning from initializer
    }
    ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._Storage.swift:24:20: error: 'self' used before 'self.init' call or assignment to 'self'
      self._buffer = _buffer
                   ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._Storage.swift:25:5: error: 'self.init' isn't called on all paths before returning from initializer
    }
    ^

/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._Storage.swift:24:20: error: 'self' used before 'self.init' call or assignment to 'self'
      self._buffer = _buffer
                   ^
/Users/srdan/dev/swift-collections/Sources/DequeModule/Deque._Storage.swift:25:5: error: 'self.init' isn't called on all paths before returning from initializer
```

### Checklist

### DIFF
--- a/Sources/DequeModule/Deque+Sequence.swift
+++ b/Sources/DequeModule/Deque+Sequence.swift
@@ -16,6 +16,7 @@ extension Deque: Sequence {
   // conversions from indices to storage slots.
 
   /// An iterator over the members of a deque.
+  @frozen
   public struct Iterator: IteratorProtocol {
     @usableFromInline
     internal var _storage: Deque._Storage
--- a/Sources/DequeModule/Deque._Storage.swift
+++ b/Sources/DequeModule/Deque._Storage.swift
@@ -10,6 +10,7 @@
 //===----------------------------------------------------------------------===//
 
 extension Deque {
+  @frozen
   @usableFromInline
   struct _Storage {
     @usableFromInline
--- a/Sources/DequeModule/Deque._UnsafeHandle.swift
+++ b/Sources/DequeModule/Deque._UnsafeHandle.swift
@@ -10,6 +10,7 @@
 //===----------------------------------------------------------------------===//
 
 extension Deque {
+  @frozen
   @usableFromInline
   internal struct _UnsafeHandle {
     @usableFromInline
--- a/Sources/DequeModule/_DequeBuffer.swift
+++ b/Sources/DequeModule/_DequeBuffer.swift
@@ -9,6 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+@_fixed_layout
 @usableFromInline
 internal class _DequeBuffer<Element>: ManagedBuffer<_DequeBufferHeader, Element> {
   @inlinable
--- a/Sources/DequeModule/_UnsafeWrappedBuffer.swift
+++ b/Sources/DequeModule/_UnsafeWrappedBuffer.swift
@@ -9,6 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+@frozen
 @usableFromInline
 internal struct _UnsafeWrappedBuffer<Element> {
   @usableFromInline
@@ -51,6 +52,7 @@ internal struct _UnsafeWrappedBuffer<Element> {
   internal var count: Int { first.count + (second?.count ?? 0) }
 }
 
+@frozen
 @usableFromInline
 internal struct _UnsafeMutableWrappedBuffer<Element> {
   @usableFromInline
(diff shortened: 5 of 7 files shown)


lorentey:
Building for distribution is not a simple decision -- it includes ABI compatibility promises that the implementations in this package aren't yet prepared to make.

Can you tell me why you need to distribute this package in binary form?

srdanrasic:
Thanks for your reply @lorentey. 

What I need is a binary, an xcframework, without ABI compatibility promise. Unfortunately, as far as I know, building a binary has an implicit requirement of building for distribution, i.e. it needs flag BUILD_LIBRARY_FOR_DISTRIBUTION to be set.

We have two use cases:

1) We are delivering a framework in a binary format that has Swift Collections as a dependency, with no ABI guarantee.
2) We have a testing setup that makes use of prebuilt packages for efficiency.

lorentey:
> We have two use cases:
> 
> 1. We are delivering a framework in a binary format that has Swift Collections as a dependency, with no ABI guarantee.

This doesn't seem like a good idea to me. You're pinning your clients to the particular version of swift-collections that you're building with, and you're also forcing them to use it in a build configuration that we aren't testing.

I believe the way to use packages in a framework that's distributed in binary form is to vendor them -- i.e., to locally duplicate their source in an internal module that uses a different name. The fact that you're using swift-collections should not interfere with your clients' ability to use it as a package.

> 2. We have a testing setup that makes use of prebuilt packages for efficiency.

This is reasonable, as long as you're okay with testing the code in a form that we aren't testing or optimizing for.

srdanrasic:
I'll go an close this PR since it has diverged from the original idea.

Issue #3, opened 2021-04-05
Priority queue


lorentey:
[Priority queues][priority queue] are commonly requested data structures, and are universally useful. A high-quality implementation would be a natural candidate for inclusion in this package. (Priority queues are used in sorting algorithms, various graph algorithms, networking protocol implementations, etc. etc. etc.)

[priority queue]: https://en.wikipedia.org/wiki/Priority_queue

I expect a straightforward array-backed binary heap might make most sense for the initial implementation, but I wouldn't be surprised if a fancier variant proved better overall.

```swift
struct Heap<Element: Comparable>
```

This would likely be mostly an implementation exercise rather than an API design challenge (although API design is *never* trivial). We'd probably need to experiment with several different data structure implementations, and perhaps even come up with our own variants that's better suited to Swift's unique combination of language features.

julianschiavo:
This would be super useful for my current use case! Sorry if this is a beginner question here, but I was wondering, what advantage would this offer over using an OrderedSet and sorting it by the priority in a model (as in, I have a model with a priority property and I sort the OrderedSet by that property)? (Also, would the pqueue auto sort upon insertion?)

634750802:
@julianschiavo It takes less computing while you read the biggest few elements. It is more effective when you only use some of the biggest elements but not all collection. It's a `queue` but not a `random access collection`.

>  what advantage would this offer over using an OrderedSet and sorting it by the priority in a model

A priority queue is not a `set`, it may allows duplicated elements (depends on implementation).

> Also, would the pqueue auto sort upon insertion?

A heap is not stored as sorted sequence, it only guarantees the top elements are bigger or smaller than bottom elements.

Heap is a complete binary tree and could be stored in a continuous array. Index `n`'s parent is `n/2` and it's children are  `n*2 + 1` and `n*2 + 2`. The heap:

```
        10
   9           3
8     7     _     _
```

It can be stored as array:

```
10 9 3 8 7 _ _
```

It is partial sorted bu not full sorted upon insertion or deletion. So it take less time to insert or delete an element.

- **Insertion**: Add a new element at the tail of the array, and `pop` the element to a proper position.

- **Deletion**: You can only remove the top element. Every removed element should replaced by the biggest direct child.

Both operation takes no more than logN times swap operation.

julianschiavo:
Got it, thank you so much for your in-depth explanation, really helps with understanding the concept! That's exactly what I'm looking for if I have that right 😄 

[...]

PR #43, opened 2021-05-25
Priority queue implementation

AmanuelEphrem:
## Purpose
This PR adds the priority queue data structure into the swift collections, as a response to issue #3. The uses for priority queues are expansive, as they are used in graph searching algorithms, networking algorithms, sorting algorithms, data compression, and much more. Including this data structure in the swift collections would accelerate developer productivity, as they wouldn’t be bogged down by implementation details, and ensure an efficient implementation. 

## Details
This implementation uses an array-based binary heap as the underlying storage, to take advantage of its O(log(n)) time complexity for `enqueue()` and `dequeue()` operations. It enforces that all elements of the `PriorityQueue`  conform to the Comparable protocol so that priority between elements can be determined. While each element could have an associated priority, this implementation would be memory intensive, as @lorentey has previously stated. I believe this drawback is too costly, especially for its initial implementation, and is why I chose against writing it this way (for now at least).  

## Incomplete Tasks
Currently, I have implemented the `PriorityQueue`  functions, added tests for each of them, and included a markdown page.  Tasks that still need to be accomplished include:
* Benchmark testing
* General testing (the tests I have included may not be comprehensive enough)

## Future Tasks
* Adding conformity to the CustomStringConvertible protocol
* Adding conformity to the Equitable protocol

Evidently this is a work-in-progress PR and would greatly appreciate both feedback and help in completing the Incomplete Tasks 😃

## Checklist


lorentey:
Thanks for implementing this! As discussed in #3, we'd like to implement a min-max heap. Would you be interested in working on that?

This variant could still be useful as a reference implementation to validate we got the expected performance results.

Sources/PriorityQueue/PriorityQueue+Sequence.swift
```diff
@@ -0,0 +1,45 @@
...
+// This source file is part of the Swift Collections open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+extension PriorityQueue:Sequence{
```
lorentey, line 12: Nit: please follow the coding style established in the rest of this project. We like to put spaces around punctuation characters like `:` and infix operators, and before opening braces. Keeping this consistent helps readability.

Sources/PriorityQueue/PriorityQueue+Sequence.swift
```diff
@@ -0,0 +1,45 @@
...
+    
+    /*
+     *  Defines an iterator to conform to Sequence
+     */
+    public struct Iterator:IteratorProtocol{
+        //the priority queue at a specific moment
+        private var snapshotPriorityQueue:PriorityQueue
```
lorentey, line 19: This property does not need to be mutable.

The name `snapshotPriorityQueue` is needlessly verbose -- I recommend something simple like `_base` instead.

```suggestion
        private let _base: PriorityQueue
```

You do not need to emphasize that this property contains a priority queue at a specific moment. That's true for every variable.

Sources/PriorityQueue/PriorityQueue+Sequence.swift
```diff
@@ -0,0 +1,45 @@
...
+        
+        public init(snapshotPriorityQueue:PriorityQueue){
```
lorentey, line 21: Such initializers do not need to be public. 

```suggestion
        internal init(_base: PriorityQueue) {
```

Sources/PriorityQueue/PriorityQueue.swift
```diff
@@ -0,0 +1,227 @@
...
+    public init(priorityQueueType:HeapType){
+        storage = [Element]()
+        storageType = priorityQueueType
+    }
+    
+    /*
+     *  Adds an element to the Priority Queue
+     *  @param  value  the element to add to the Priority Queue
+     */
+    public mutating func enqueue(value:Element){
```
lorentey, line 40: For `Deque`, we decided to go with a naming scheme that fits better with the existing operations in the stdlib: instead of `enqueue`, we have `append`; instead of `dequeue` we have `removeFirst`/`popFirst`.

I think we'll want to use a similar approach here -- instead of `enqueue`/`dequeMin`/`dequeMax`, we'd go with something like `insert`/`removeMin`/`removeMax`.

Sources/PriorityQueue/PriorityQueue.swift
```diff
@@ -0,0 +1,227 @@
...
+        storage.append(value)
+        swimUp(elementIndex: storage.count-1);
+    }
+    
+    /*
+     *  Moves a specified element down the array (towards index zero) until it's in its natural position
+     *  @param  elementIndex  the index of the specified element
+     */
+    private mutating func swimUp(elementIndex:Int){
```
lorentey, line 49: This package follows [The Leading Underscore Rule](https://github.com/apple/swift/blob/main/docs/StandardLibraryProgrammersManual.md#the-leading-underscore-rule) from the stdlib naming conventions; internal or private members need to have a leading underscore somewhere in their fully qualified name.

```suggestion
    private mutating func _raise(from index: Int) {
```

Sources/PriorityQueue/PriorityQueue.swift
```diff
@@ -0,0 +1,227 @@
...
+        }
+        
+        
+    }
+    
+    /*
+     *  Determines the element with the highest/lowest priority, depending on the heap type
+     *  @return  the element with the highest/lowest priority, or nil if no elements exist
+     */
+    public func peek() -> Element?{
```
lorentey, line 189: I think it would be better to reuse the `min()`/`max()` member names from the `Sequence` protocol.

Sources/PriorityQueue/PriorityQueue.swift
```diff
@@ -0,0 +1,227 @@
...
+            return nil;
+        }
+        return storage[0];
+    }
+    
+    /*
+     *  Determines whether the Priority Queue is full.
+     *  @return  true if the PriorityQueue is full, false otherwise
+     */
+    public func isFull() -> Bool{
```
lorentey, line 200: We don't need this.

Sources/PriorityQueue/PriorityQueue.swift
```diff
@@ -0,0 +1,227 @@
...
+     *  @return  true if the Priority Queue is empty, false otherwise
+     */
+    public func isEmpty() -> Bool{
+        return size() == 0;
+    }
+    
+    /*
+     *  Determines the size of the Priority Queue
+     */
+    public func size() -> Int{
```
lorentey, line 215: This needs to be a property called `count`.

```suggestion
    public var count: Int {
```

AmanuelEphrem:
@lorentey, thank you for the fast reply and comments. I seemed to have jumped the gun and went straight into the priority queue, so I can easily convert the PriorityQueue.swift file into a general min/max heap implementation. 

It looks like @AquaGeek and I both accidentally implemented (more or less) the same min/max heap, so we can also find ways to consolidate both our implementations into one. I think if we keep my PriorityQueue.swift file (going to change it to Heap.swift) and then keep @AquaGeek 's PriorityQueue.swift as a wrapper over Heap.swift that can work nicely. Currently I am forcing the user to choose between a min or max heap, but @AquaGeek implemented (wonderfully) a min and max heap. So this is a place where we can "merge" @AquaGeek 's additional functions with my Heap.swift file. This way, conformance to the Sequence protocol would not be effected. 

Please let me know how this method to consolidate both implementations look, and feel free to suggest a new way of doing it. This is just my initial thoughts, so any and all feedback would be great!

AquaGeek:
> It looks like @AquaGeek and I both accidentally implemented (more or less) the same min/max heap, so we can also find ways to consolidate both our implementations into one.

I'd be happy to work together to consolidate our two implementations!

Sources/PriorityQueue/PriorityQueue.swift
AmanuelEphrem: I noted to change this in the comments, once Heap.swift allows for retrieving both the min **and** max values, instead of just one of them.

PR #44, opened 2021-05-27
[Deque] Work around stdlib issue with Array.withContiguousStorageIfAvailable

### LINKED ISSUE #27: Fatal error: unsupported: file Swift/ArrayBuffer.swift, line 231
When initializing a `Deque` from an `Array` (coming from `NSArray`), it crashes at runtime with error:

```
Fatal error: unsupported: file Swift/ArrayBuffer.swift, line 231
```

### Information

- **Package version:** commit 65a3841
- **Platform version:** macOS 11.2.1 (20D74)
- **Swift version:** Apple Swift version 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)
Target: x86_64-apple-darwin20.3.0

### Checklist

### Steps to Reproduce

Crashes at runtime (Deque+RangeReplaceableCollection.swift:144):
```swift
let array = NSArray(arrayLiteral: NSString()) as! [NSObject]
let deque = Deque(array)
```

Runs without crashing:
```swift
let array = NSArray(arrayLiteral: NSString()) as! [NSObject]
let deque = Deque(Array(array))
```

### Expected behavior
The first example above doesn’t crash.

### Actual behavior
Only the second example doesn’t crash.

lorentey:
Resolves #27 by updating `Deque.init(_:)`, `.append(contentsOf:)` and `.prepend(contentsOf:)` to not call `Array.withContiguousStorageIfAvailable` if there is a chance doing so may trigger the underlying stdlib issue.

### Checklist

### DIFF
--- a/Sources/DequeModule/Compatibility.swift
+++ b/Sources/DequeModule/Compatibility.swift
@@ -1,0 +1,62 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Collections open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+extension Array {
+  /// Returns true if `Array.withContiguousStorageIfAvailable` is broken
+  /// in the stdlib we're currently running on.
+  ///
+  /// See https://bugs.swift.org/browse/SR-14663.
+  @inlinable
+  internal static func _isWCSIABroken() -> Bool {
+    #if _runtime(_ObjC)
+    guard _isBridgedVerbatimToObjectiveC(Element.self) else {
+      // SR-14663 only triggers on array values that are verbatim bridged
+      // from Objective-C, so it cannot ever trigger for element types
+      // that aren't verbatim bridged.
+      return false
+    }
+
+    // SR-14663 was introduced in Swift 5.1. Check if we have a broken stdlib.
+
+    // The bug is caused by a bogus precondition inside a non-inlinable stdlib
+    // method, so to determine if we're affected, we need to check the currently
+    // running OS version.
+    #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+    guard #available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, *) else {
+      // The OS is too old to be affected by this bug.
+      return false
+    }
+    #endif
+    // FIXME: When a stdlib is released that contains a fix, add a check for it.
+    return true
+
+    #else
+    // Platforms that don't have an Objective-C runtime don't have verbatim
+    // bridged array values, so the bug doesn't apply to them.
+    return false
+    #endif
+  }
+}
+
+extension Sequence {
+  // An adjusted version of the standard `withContiguousStorageIfAvailable`
+  // method that works around https://bugs.swift.org/browse/SR-14663.
+  @inlinable
+  internal func _withContiguousStorageIfAvailable_SR14663<R>(
+    _ body: (UnsafeBufferPointer<Element>) throws -> R
+  ) rethrows -> R? {
+    if Self.self == Array<Element>.self && Array<Element>._isWCSIABroken() {
+      return nil
+    }
+
+    return try self.withContiguousStorageIfAvailable(body)
+  }
+}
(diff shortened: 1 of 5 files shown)


Sources/DequeModule/Deque+Extras.swift
```diff
@@ -166,6 +166,12 @@ extension Deque {
   /// - SeeAlso: `append(contentsOf:)`
   @inlinable
   public mutating func prepend<S: Sequence>(contentsOf newElements: S) where S.Element == Element {
+    let done: Void? = newElements._withContiguousStorageIfAvailable_SR14663 { source in
```
lorentey, line 169: I couldn't resist slipping in this trivial performance improvement, if only for symmetry with the sequence-based `append(contentsOf:)`.

lorentey:
@swift-ci test

Sources/DequeModule/Compatibility.swift
```diff
@@ -0,0 +1,60 @@
...
+
+extension Sequence {
+  // An adjusted version of the standard `withContiguousStorageIfAvailable`
+  // method that works around https://bugs.swift.org/browse/SR-14663.
+  @inlinable
+  internal func _withContiguousStorageIfAvailable_SR14663<R>(
+    _ body: (UnsafeBufferPointer<Element>) throws -> R
+  ) rethrows -> R? {
+    if Self.self == Array<Element>.self && Array<Element>._isWCSIABroken() {
+      return nil
```
lorentey, line 55: If we had a reliable/easy way to test if a particular array instance is bridged, then we could let native arrays pass through here. 🤔

lorentey:
@swift-ci test

Tests/DequeTests/DequeTests.swift
```diff
@@ -319,5 +320,55 @@ final class DequeTests: CollectionTestCase {
...
+              prependSequence(contentsOf: extra, to: &deque)
+              expectEqualElements(deque, contents)
+            }
+          }
+        }
+      }
+    }
+  }
+
+  func test_prependManyFromBridgedArray() {
```
stephentyrone, line 351: Thank you for adding the additional test cases =)

stephentyrone (approved):

Sources/DequeModule/Compatibility.swift
karwa: Maybe:

```swift
if let array = self as? Array<Element>, array._baseAddressIfContiguous == nil { return nil }
```

Tested in a Playground:
```swift
import Foundation
var a: [NSObject] = [NSObject()]
print(a._baseAddressIfContiguous) // Optional(0x00007fdf7e40c0d0)

a = NSArray(array: [NSObject()]) as! [NSObject]
print(a._baseAddressIfContiguous) // nil
```

lorentey: Huh, that's a good find. Looking at similarly exposed members, `Array._hoistableIsNativeTypeChecked()` might be even better.

There are (at least) two worries with using these underscored APIs: (including `_isBridgedVerbatimToObjectiveC` above)

1. Swift doesn't promise it won't change their behavior, so it's tricky to safely use them in code that is not part of the stdlib. E.g., I imagine for element types that don't require a dynamic type check at every access, we may want to change `_baseAddressIfContiguous` to return non-nil for bridged arrays that use specific known NSArray subclasses. 

2. These have also been made public for historical reasons, rather than any intentional design, and we may find a good reason to change them to `@inlinable internal`, disallowing direct calls like this in future SDKs. We wouldn't be able to remove the entry points themselves (they are part of Swift's ABI), but we might need to restrict them to only be called from inlinable functions within the stdlib.

As this package is maintained by the same people who work on the Swift stdlib, I think it's not unreasonable for this package to use these. At the first sign of trouble, we can remove the calls, or limit them to older Swift toolchains. The primary danger is that this code will spread beyond this package through vendoring or uncritical copy-n-paste. 

My instinct as a stdlib engineer tells me that the risk is low with `_hoistableIsNativeTypeChecked` (and pretty much nonexistent for `_isBridgedVerbatimToObjectiveC`); on the other hand, this package is inherently averse to even low-risk stuff. Hm.

lorentey: I think the best thing to do is to keep this as is for now until we have a Swift release (let's call it x.y) that contains a fix for the underlying issue. In the (unlikely) case that the fix won't back deploy, then we can add a call to `_hoistableIsNativeTypeChecked` restricted to only be called on Swift 5.1 ..< x.y.

stephentyrone: I think that's fine. I would also be OK with doing the workaround here now, so long as we put a comment in the stdlib that references this bug.

PR #51, opened 2021-05-28
Add a priority queue implementation built on top of `Heap`

AquaGeek:
### Description

This introduces a double-ended `PriorityQueue` as discussed in #3.

### Detailed Design

`Heap` (a min-max heap) is used as the storage. One of the big advantages of using a min-max heap is that we don't need to bifurcate into a min heap or max heap and keep track of which kind is used — you can pop min or max, as needed.

The main difference between `Heap` and `PriorityQueue` is that the latter separates the value from the comparable priority of it. This is useful in cases where a type doesn't conform to `Comparable` directly but it may have a property that does — e.g. `Task.priority`. It also keeps track of insertion order, so dequeueing of elements with the same priority should happen in FIFO order.

```swift
public struct PriorityQueue<Value, Priority: Comparable> {
    public typealias Element = (value: Value, priority: Priority)

    public var isEmpty: Bool
    public var count: Int

    // Initializers
    public init()
    public init<S: Sequence>(_ elements: S) where S.Element == Element
    public init(arrayLiteral elements: Element...)
    public init(dictionaryLiteral elements: (Value, Priority)...)

    // Mutations/querying
    public mutating func insert(_ value: Value, priority: Priority)
    public func min() -> Value?
    public func max() -> Value?
    public mutating func popMin() -> Value?
    public mutating func popMax() -> Value?
    public mutating func removeMin() -> Value
    public mutating func removeMax() -> Value
}
```

### To-Do

- [x] Merge @AmanuelEphrem's `Sequence` conformance
- [x] Reformat the code to match the 2 space indent/80 char width style
- [x] Optimize `_indexOfChildOrGrandchild(of:sortedUsing:)` (probably by splitting it back out into 2 separate functions)
- [x] Chase down [suggestion from @timvermeulen](https://github.com/apple/swift-collections/pull/51#discussion_r651145630) about skipping nodes that have children
- [x] Figure out if `@inlinable` is required
- [x] Add checks for heap invariants
- [x] Make a call on whether or not we want to [add `removeMin/Max`](https://github.com/apple/swift-collections/pull/51#discussion_r651089847)
- [x] Make `_minMaxHeapIsMinLevel` an instance method
- [x] Merge @hassila's changes to make `_bubbleUpMin(startingAt:)` and `_trickleDown(startingAt:)` iterative instead of recursive
- [X] Add implementation of `insert(contentsOf:)`
- [X] Rebase this on top of the `Heap` implementation merged into `main`

### Future Directions

- Add support for merging two priority queues
- Add `replaceMin` and `replaceMax`

### Documentation

The public APIs have largely been documented. An overview document has been added to the Documentation directory.

### Testing

There are unit tests for the added `PriorityQueue` type.

### Performance

Performance tests have been added. @lorentey has a PR up (#76) which adds a `std::priority_queue` benchmark — it would be interesting to compare against that with more complex types.

We may want to revisit the library JSON to ensure we have the desired comparisons defined.

### Source Impact

This is purely additive. No existing APIs were changed, deprecated, or removed.

### Checklist
- [X] I've added tests covering all new code paths my change adds to the project (to the extent possible).
- [X] I've verified that my change does not break any existing tests or introduce unexpected benchmark regressions.
- [X] I've updated the documentation (if appropriate).


kylemacomber:
This looks really good!

@timvermeulen Could you lend a second pair of eyes and review the heap logic?

> The `PriorityQueue` type is a very thin wrapper on top of `MinMaxHeap`. I went through and renamed some of the functions in `MinMaxHeap` based on feedback @lorentey provided in #43; after doing so, it became an even thinner wrapper. Maybe it's not even needed?

I agree. I don't think the separate `MinMaxHeap` type is needed. Let's just have a single type called `PriorityQueue`!

It might be nice to add a few convenience APIs for bulk insertion/creation:
- `public init<S: Sequence>(_ elements: S) where S.Element == Element`
- `public mutating func insert<S: Sequence>(contentsOf newElements: S) where S.Element == Element`
- `ExpressibleByArrayLiteral` conformance

> I have not yet run the performance tests to ensure the performance matches what is expected.

I think this would be the most profitable next step to take. It'd be great to see how the performance of `PriorityQueue` compares to `CFBinaryHeap`!

hassila:
Super to see this in progress! Just one quick initial comment/question while browsing, push/pop is documented as `Complexity: O(log n)` - although if skipping levels [as mentioned in article referenced previously in #3](https://probablydance.com/2020/08/31/on-modern-hardware-the-min-max-heap-beats-a-binary-heap/) the time complexity should be `O(log n) / 2` - different implementation or documentation?

AquaGeek:
> Just one quick initial comment/question while browsing, push/pop is documented as `Complexity: O(log n)` - although if skipping levels [as mentioned in article referenced previously in #3](https://probablydance.com/2020/08/31/on-modern-hardware-the-min-max-heap-beats-a-binary-heap/) the time complexity should be `O(log n) / 2` - different implementation or documentation?

The implementation does skip levels (see `_bubbleUpMin` and `_bubbleUpMax` — they reach for the grandparent nodes). My understanding is that when using big-O notation coefficients are usually not included.

hassila:
Ok, thanks for clarification!

AquaGeek:
In the case of `Sequence` conformance, how do we handle the double-endedness of the `PriorityQueue`? i.e. Should `next()` iterate from lowest to highest priority or vice-versa?

kylemacomber:
> In the case of `Sequence` conformance, how do we handle the double-endedness of the `PriorityQueue`? i.e. Should `next()` iterate from lowest to highest priority or vice-versa?

Maybe `PriorityQueue` shouldn't conform directly to `Sequence` but have (warning placeholder names!) a `minHeap` and `maxHeap` "view":

```swift
let q: PriorityQueue = ...

for elt in q.minHeap {
  // iterates from min to max
}

for elt in q.maxHeap {
  // iterates from max to min
}
```

wdyt?

AquaGeek:
> > In the case of `Sequence` conformance, how do we handle the double-endedness of the `PriorityQueue`? i.e. Should `next()` iterate from lowest to highest priority or vice-versa?
> 
> Maybe `PriorityQueue` shouldn't conform directly to `Sequence` but have (warning placeholder names!) a `minHeap` and `maxHeap` "view":
> 
> ```swift
> let q: PriorityQueue = ...
> 
> for elt in q.minHeap {
>   // iterates from min to max
> }
> 
> for elt in q.maxHeap {
>   // iterates from max to min
> }
> ```
> 
> wdyt?

Another approach I was mulling over is exposing separate functions for creating the iterator (e.g. `makeLowToHighPriorityIterator` and `makeHighToLowPriorityIterator`). That doesn't address the issue that will come up if somebody tries to iterate directly (though I guess that could be addressed by not conforming to `Sequence` directly).

I think exposing a view that conforms to `Sequence` makes more sense.

hassila:
https://developer.apple.com/documentation/swift/bidirectionalcollection ?

kylemacomber:
> https://developer.apple.com/documentation/swift/bidirectionalcollection ?

If we can provide a bidirectional collection view I think that'd work great:

```swift
for elt in q.sortedElements {
  // iterates from min to max
}

for elt in q.sortedElements.reversed() {
  // iterates from max to min
}
```

However, it wasn't clear to me if we're able to provide an efficient bidirectional view over the heap.

AmanuelEphrem:
> If we can provide a bidirectional collection view I think that'd work great:

Would the priority queue be able to conform to the Collection protocol? Because min and max values are determined during the remove operation, the priority queue does not lend itself easily to random access of elements via indexes.

>```swift
>let q: PriorityQueue = ...
>
>for elt in q.minHeap {
>  // iterates from min to max
>}
>
>for elt in q.maxHeap {
 > // iterates from max to min
>}```

I like this method of traversing the priority queue, as it is the least ambiguous. 

>That doesn't address the issue that will come up if somebody tries to iterate directly

In this case, the iterator could default to iterating from min to max (or max to min if that makes more sense)

hassila:
I don’t think the iterator necessarily should default to any sorted order - it could just provide “some” order unless asked for sorted / reversed  as suggested? Just trying to find reuse of existing protocols if possible if performant and natural (that being said, can’t say whether efficiency would be good - but just wanted to point out that there is an existing interface that might be useful) - that would also argue for popFirst/popLast methods  for dequeueing from respective ends

kylemacomber:
> Because min and max values are determined during the remove operation, the priority queue does not lend itself easily to random access of elements via indexes.

This was my impression.

> I don’t think the iterator necessarily should default to any sorted order - it could just provide “some” order unless asked for sorted / reversed as suggested?

I think both could be useful:

```swift
let q: PriorityQueue<Foo> = ...
let heap: [Element] = q.heap // direct access to the underlying heap array
for elt in q.minToMax { ... } // a sequence that iterates from min to max
for elt in q.maxToMin { ... } // a sequence that iterates from max to min
```

Direct access to the underlying heap array could be useful if want to efficiently hand off all the elements to an API that expects an Array if order isn't important.

AmanuelEphrem:
> Direct access to the underlying heap array could be useful if want to efficiently hand off all the elements to an API that expects an Array if order isn't important

I feel that exposing the underlying heap to the developer would introduce unnecessary complexity, and a separate function could return an array view of the priority queue.

Also, because elements of the highest priority are at the front of the priority queue, another way to conform to the Sequence protocol would be like this

```swift
let q: PriorityQueue = ...

for elt in q {
 // iterates from max to min
}

for elt in q.reversed() {
// iterates from min to max
}
```
While I think this is the most straightforward way to traverse the priority queue, I'm not sure how intuitive it is for others.

AquaGeek:
I think conforming to `Collection` is out — I don't think there's a way to do this efficiently.

> > Direct access to the underlying heap array could be useful if want to efficiently hand off all the elements to an API that expects an Array if order isn't important
> 
> I feel that exposing the underlying heap to the developer would introduce unnecessary complexity, and a separate function could return an array view of the priority queue.

I think we should expose it as `var unorderedElements: [Element]`.

> Also, because elements of the highest priority are at the front of the priority queue[…]

Because we're using a min-max heap, the lowest-priority element is at the front of the backing array not the highest.

> […] another way to conform to the Sequence protocol would be like this
> 
> ```swift
> let q: PriorityQueue = ...
> 
> for elt in q {
>  // iterates from max to min
> }
> 
> for elt in q.reversed() {
> // iterates from min to max
> }
> ```
> 
> While I think this is the most straightforward way to traverse the priority queue, I'm not sure how intuitive it is for others.

I'm still in the camp of not adding conformance to `Sequence` to the priority queue type itself; I think that conformance belongs on the min and max views. I don't have good suggestions as far as naming those, though. I think calling them `minHeap` and `maxHeap` exposes some of the implementation details.

Meta question: does this sort of discussion belong in the forums or would we rather it happen here?

AmanuelEphrem:
> I think that conformance belongs on the min and max views

I can get behind this idea as well!

> I don't have good suggestions as far as naming those, though. I think calling them minHeap and maxHeap exposes some of the implementation details.

Maybe have something like this? The priority queue would not conform to the Sequence protocol, yet rather have whatever `q.orderedIncreasing()` and `q.orderedDecreasing()` returns to conform to Sequence.

```swift
 let q: PriorityQueue = ...
 
 for elt in q.orderedIncreasing() {
   // iterates from min to max
 }
 
 for elt in q.orderedDecreasing() {
   // iterates from max to min
 }
```

Assuming we are exposing the underlying heap as `var unorderedElements: [Element]`, the naming of `q.orderedIncreasing()` and `q.orderedDecreasing` would follow in the same fashion.

For brevity, however, it might be better just to have it as `q.increasing()` and `q.decreasing()` as saying "increasing"/"decreasing" already implies that it is ordered.

> Meta question: does this sort of discussion belong in the forums or would we rather it happen here?

I think discussion here is better, as we would be able to see code changes and implementation suggestions all in one place.

pyrtsa:
Other possible naming options: `ascendingElements`/`descendingElements`, or `increasingElements`/`decreasingElements`. I think the former of these two could go better together with the Foundation type `ComparisonResult` with its cases `orderedAscending` etc.

kylemacomber:
> Other possible naming options: `ascendingElements`/`descendingElements`, or `increasingElements`/`decreasingElements`. I think the former of these two could go better together with the Foundation type `ComparisonResult` with its cases `orderedAscending` etc.

I also like riffing off of `ascending` and `descending`. I think we can get away with dropping the "elements" suffix and parenthesis. For example, `OrderedSet` has an `unordered` view.

That would leave us with:

```swift
let q: PriorityQueue<Element> = ...
let heap: [Element] = q.unordered // direct read-only access to the underlying heap array
for elt in q.ascending { ... } // a sequence that iterates from min to max
for elt in q.descending { ... } // a sequence that iterates from max to min
```

I think I still prefer `heap` to `unordered` because it provides additional useful information (the elements aren't arbitrarily unordered like a hash table). However, this seems to me like a solid direction regardless of the names. I think @AquaGeek, you should proceed with your preferred names.

>> Meta question: does this sort of discussion belong in the forums or would we rather it happen here?
>
> I think discussion here is better, as we would be able to see code changes and implementation suggestions all in one place.

I think this discussion is working well in this thread—we seem to be making progress and converging. If we find ourselves at a stalemate, that might be a good time to gather more opinions by bringing it to the forums. And before we merge the PR, it'd prob be good to solicit feedback and do some kind of API review on the forums.

AquaGeek:
@AmanuelEphrem Do you want to work on implementing `ascending` and `descending` and open a PR against my branch? I was thinking of returning an `Iterator` from both of those that conforms to both `IteratorProtocol` **and** `Sequence`:

```swift
extension PriorityQueue {
    public struct Iterator: IteratorProtocol, Sequence {
        private var _base: PriorityQueue
        private let direction: IterationDirection

        […]
    }

    var ascending: Iterator {
        Iterator(_base: Self, direction: .ascending)
    }

    var descending: Iterator {
        Iterator(_base: Self, direction: .descending)
    }
}
```

I'll work on adding the performance tests.

AquaGeek:
Here's a first pass at performance tests:

chart-pq

Benchmarks/Benchmarks/PriorityQueueBenchmarks.swift
```diff
@@ -0,0 +1,64 @@
...
+                blackHole(queue)
+            }
+        }
+
+        self.add(
+            title: "PriorityQueue<Int> removeMin",
+            input: [Int].self
+        ) { input in
+            return { timer in
+                var queue = PriorityQueue(input)
```
AquaGeek, line 55: Should this go above the closure (so that we're not also timing the instantiation)?

AmanuelEphrem: I think it should still stay within the outer closure, but outside the inner closure (essentially swapping lines 54 and 55) so it stays consistent with the other benchmarks.

hassila:
> Here's a first pass at performance tests:
> 
> 
> 
> chart-pq

Interesting! Out of curiosity what hardware is this running on? (And to confirm it is built in release mode?)

AquaGeek:
> Interesting! Out of curiosity what hardware is this running on? (And to confirm it is built in release mode?)

This was run on a 2019 MacBook Pro (2.3 GHz 8-Core Intel Core i9). Yes, it was built in release mode.

kylemacomber:
> Here's a first pass at performance tests:
> 
> chart-pq

@AquaGeek Nice! It'd be great to see how the performance compares to [CFBinaryHeap](https://developer.apple.com/documentation/corefoundation/cfbinaryheap-s11) for equivalent operations (e.g. `CFBinaryHeapAddValue` and `CFBinaryHeapGetMinimum`)!

AquaGeek:
> @AquaGeek Nice! It'd be great to see how the performance compares to [CFBinaryHeap](https://developer.apple.com/documentation/corefoundation/cfbinaryheap-s11) for equivalent operations (e.g. `CFBinaryHeapAddValue` and `CFBinaryHeapGetMinimum`)!

@kylemacomber Here's a quick pass at adding `CFBinaryHeap` implementations. I'll get the code added here shortly. Somebody _definitely_ needs to check my work here — take the benchmarks with a hefty grain of salt.

CFBinaryHeap Comparison

hassila:
@AquaGeek - just spent a little time doing an initial profiling, as I got similar results as you on my own machine - the run times per element seemed long compared to the C++ results in the article references earlier so I wanted to dig a bit on root causes. I found at least one thing which is a bummer - it seems swapAt will hit malloc [and it is a known issue](https://forums.swift.org/t/unexpected-swapat-free-malloc-overhead/41109/2). So it seems to use around 30% of total run time goes there.

hassila:
I just spent a little time profiling a bit (as the benchmark runtime seems quite long compared to the C++ version [referenced in the article linked to previously](https://probablydance.com/2020/08/31/on-modern-hardware-the-min-max-heap-beats-a-binary-heap/)), there seems to be an issue with using swapAt: (probably the same [issue discussed here](https://forums.swift.org/t/unexpected-swapat-free-malloc-overhead/41109)) - I just did a simple measurement of `PriorityQueue<Int> insert` and could see around 7.5M transient allocations. Doing the simple change of manually doing the swapAt: like this in all places:

```
storage.swapAt(largestDescendantIdx, index)
```

changed to

```
let tmp = storage[index]
storage[index] = storage[largestDescendantIdx]
storage[largestDescendantIdx] = tmp
```

brought this test down to 165K transient allocations and the runtime was basically halved.

And the original benchmark went from:
chart-old

to:

chart-new

Not sure if this is a known issue with the swapAt: method, but it seems a bit rough to force two memory allocations per swapAt: operation for a simple small value type.

Also, looking at the removeMax test, there is also a large amount of transient allocations with this backtrace, can't quite understand why though (haven't looked at any others yet).

image

hassila:
Tracked down the removeMax transient allocations too (disable inlining of _trickleDownMax and used Instruments), removing them changed the runtime for one iteration of the benchmark from from 95s -> 83s (and removed several M allocations too).

It was triggered by a couple of guard statements looking like:
```
        guard let (smallestDescendantIdx, isChild) = _indexOfChildOrGrandchild(of: index, sortedUsing: <) else {
            return
        }
```

So I just tried remove 'sortedUsing' and created two duplicate methods for _indexOfChildOrGrandchild (one for greater, one for less than) and the allocations went away (it seems an implicit closure is created by passing the operator function as an argument which caused all the allocations).

image

glessard:
> storage.swapAt(largestDescendantIdx, index)

I wonder if `storage.withUnsafeMutableBufferPointer { $0.swapAt(largestDescendantIdx, index) }` would also exhibit the excessive allocations problem?

timvermeulen (commented):
This looks great, @AquaGeek! Min-max heaps are awesome, thanks for providing this implementation. I've left some suggestions based on things I noticed when playing around with the code.

Two other things:
- The rest of this package uses 2-space indentation and a line width of 80, please update your PR to match this.
- Since heaps have some invariants that need to be maintained (min-max heaps especially), you could consider adding a method that checks these invariants just for testing purposes. With some of the operations it can be a bit tricky to keep a clear mental model of exactly what's happening.

Thanks also for already providing some basic benchmarks, they made experimenting with changes a lot easier 🙂

hassila:
> > storage.swapAt(largestDescendantIdx, index)
> 
> I wonder if `storage.withUnsafeMutableBufferPointer { $0.swapAt(largestDescendantIdx, index) }` would also exhibit the excessive allocations problem?

I gave it a spin, and it does not have the transient allocations - it does perform significantly worse though at 7.7s instead of ~6s for the manually inlined version of swapAt (vs ~12s with the transient allocations using swapAt).

hassila:
@AquaGeek for the swapAt: issue, I tried the following with good results - add a helper method:

```
    @inline(__always)
    private mutating func _swapAt(_ first: Int, _ second: Int) {
        let tmp = self.storage[first]
        self.storage[first] = self.storage[second]
        self.storage[second] = tmp
    }
```

and just change the call site from:

`storage.swapAt(smallestDescendantIdx, index)`

to

`self._swapAt(smallestDescendantIdx, index)`

hassila:
I created https://bugs.swift.org/browse/SR-14778 for the swapAt: issue.

glessard:
> it does perform significantly worse though at 7.7s instead of ~6s for the manually inlined version of swapAt

Interesting. I would think that the manually-inlined would do worse than the UMBP-calling version when used on types that do ARC, such as String. The specialization of `swapAt` on UMBP exists partly to explicitly avoid ARC traffic, and to not rely on the optimizer to remove it. It looks to me like the optimizer still can't reliably do that.

Looking at the generated code of the following on godbolt:
```swift
var a = [1,2,3,4,5]
var b = a.map(String.init)
var c = b
var d = b

extension Array {
    @inlinable @inline(__always)
    public mutating func _swapAt(_ first: Int, _ second: Int) {
        let tmp = self[first]
        self[first] = self[second]
        self[second] = tmp
    }
}

@inline(never)
func sa(_ a: inout [Int]) {
    a._swapAt(0, 3)
}

@inline(never)
func sb(_ b: inout [String]) {
    b._swapAt(0, 3)
}

@inline(never)
func sc(_ c: inout [String]) {
    c.withUnsafeMutableBufferPointer { $0.swapAt(0, 3) }
}

@inline(never)
func sd(_ d: inout [String]) {
    d.swapAt(0, 3)
}

sa(&a)
sb(&b)
sc(&c)
sc(&d)
```

`sb` and `sd` look about the same, while `sc` eliminates the retains and releases. Does it make sense to also benchmark PriorityQueue with a non-trivial type?

hassila:
Interesting. In reality I’d personally probably end up using a simple struct with priority (Int) and then an index / key to a collection type for the actual entities in the priority queue - just to ensure cheap push/pop and no arc if keeping references. But it wouldn’t be bad testing it of course, as your use-case may vary.

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,363 @@
...
+
+        return result
+    }
+
+    // MARK: - Helpers
+
+    @inline(__always)
+    private func _parentIndex(of index: Int) -> Int {
+        (index - 1) / 2
+    }
```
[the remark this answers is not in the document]
hassila: @timvermeulen - that should only be required for the public interface, [module private/internal methods should inline with just `@inline(__always)`](https://forums.swift.org/t/when-should-both-inlinable-and-inline-always-be-used/37375/2) unless I've missed something (and looking at the stack traces from Instruments, it definitely seems inlining have worked..). @inlinable would be required for the public methods though!

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,363 @@
...
+            return Swift.max(left, right)
+        }
+    }
+
+    /// Removes and returns the element with the lowest priority, if available.
+    ///
+    /// - Complexity: O(log n)
+    public mutating func removeMin() -> Element? {
+        return _delete(at: 0)
+    }
```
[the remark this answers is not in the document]
AquaGeek: I don't have strong opinions about the names here. @lorentey's [original feedback](https://github.com/apple/swift-collections/pull/43#discussion_r641254919) of @amanuelephrem's PR suggested `removeMin/Max`, so that's why I went with `remove*`. `popMin/Max` seems more in line with the end result of these methods to me, so I'm happy to tweak the naming.

Sources/PriorityQueueModule/PriorityQueue.swift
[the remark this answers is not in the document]
AquaGeek: Can somebody confirm this? Is `@inlinable` required?

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,363 @@
...
+
+/// Returns `true` if `count` elements falls on a min level in a min-max heap.
+///
+/// - Precondition: `count` must be > 0.
+@inline(__always)
+func _minMaxHeapIsMinLevel(_ count: Int) -> Bool {
+    precondition(count > 0)
+
+    return count._binaryLogarithm() & 0b1 == 0
+}
```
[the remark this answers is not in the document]
AquaGeek: I had moved this out in my original implementation that was generic around both the element and its comparable key path, because it seemed strange to have to specify the generics in the tests.

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,363 @@
...
+// MARK: -
+
+extension PriorityQueue {
+    /// Initializes a queue from a collection.
+    ///
+    /// Utilizes [Floyd's linear-time heap construction algorithm](https://en.wikipedia.org/wiki/Heapsort#Floyd's_heap_construction).
+    public init<C: Collection>(_ collection: C) where C.Element == Element {
+        storage = Array(collection)
+
+        for idx in (0...(storage.count / 2)).reversed() {
```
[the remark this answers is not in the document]
AquaGeek: Yep, you're right. If `count` is 20, we only need to iterate to 9 (as 10 is the first index of the second half of the array).

AquaGeek:
Thank you to everyone who provided feedback on this. I have integrated most of it.

Remaining tasks:
- [ ] Merge @AmanuelEphrem's `Sequence` conformance (waiting on some tweaks from his side)
- [ ] Reformat the code to match the 2 space indent/80 char width style
- [ ] Optimize `_indexOfChildOrGrandchild(of:sortedUsing:)` (probably by splitting it back out into 2 separate functions)
- [ ] Chase down [suggestion from @timvermeulen](https://github.com/apple/swift-collections/pull/51#discussion_r651145630) about skipping nodes that have children
- [ ] Figure out if `@inlinable` is required
- [ ] Add unit tests for heap invariants
- [ ] Make a call on whether or not we want to [add `removeMin/Max`](https://github.com/apple/swift-collections/pull/51#discussion_r651089847) in addition to `popMin/Max` as well as `replaceMin/Max`

hassila:
Very cool @AquaGeek, will play a bit now with the latest version! :-)

Just in case it was missed as it was part of a different comment above and now marked resolved, @timvermeulen suggestion on an iterative implementation of e.g. `mutating func _bubbleUpMin(startingAt index: Int) {` might be worth investigating too?

hassila:
@AquaGeek I found a way to keep `_indexOfChildOrGrandchild` as a single method while removing the transient allocations with minimal changes.


1. Change the signature of the method to:
```
@inline(__always)
private func _indexOfChildOrGrandchild(of index: Int, greaterThan: Bool) -> (index: Int, isChild: Bool)? {
```

The previous predicate code becomes:
```
// if predicate(storage[rightChildIdx], storage[leftChildIdx]) {
if greaterThan ? storage[rightChildIdx] > storage[leftChildIdx] : storage[rightChildIdx] < storage[leftChildIdx] {
```
and

```
// if predicate(storage[i], storage[result.index]) {
if greaterThan ? storage[i] > storage[result.index] : storage[i] < storage[result.index] {
```

And finally the two call sites changes to:
```
guard let (largestDescendantIdx, isChild) = _indexOfChildOrGrandchild(of: index, greaterThan: true) else {
guard let (smallestDescendantIdx, isChild) = _indexOfChildOrGrandchild(of: index, greaterThan: false) else {
```

This removes the transient allocations and significantly drops the execution time.

hassila:
Before/after for removeMax with that optimization:

chart-new-latest


chart-new-latest2

hassila:
FYI, I've looked a bit more on the transient allocations in removeMax now, it seems the Array triggers copy-on-write even if it shouldn't sometimes:

image

(closest issue I found was https://bugs.swift.org/browse/SR-14062, but unclear if it is the same - will dig a bit more)

AquaGeek:
> Just in case it was missed as it was part of a different comment above and now marked resolved, @timvermeulen suggestion on an iterative implementation of e.g. `mutating func _bubbleUpMin(startingAt index: Int) {` might be worth investigating too?

Ooops. I had missed that when I was making the list. Thanks for reminding me @hassila!

AquaGeek:
Here are the updated benchmarks:

pq-results

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,490 @@
...
+  /// Initializes a queue from a collection.
+  ///
+  /// Utilizes [Floyd's linear-time heap construction algorithm](https://en.wikipedia.org/wiki/Heapsort#Floyd's_heap_construction).
+  public init<C: Collection>(_ collection: C) where C.Element == Element {
+    storage = Array(collection)
+
+    for idx in (0..<(storage.count / 2)).reversed() {
+      _trickleDown(startingAt: idx)
+    }
+  }
```
kylemacomber, line 477: This linear-time initializer is awesome!

I think the algorithm over `Collection` is redundant with the more general algorithm over `Sequence` below.

kylemacomber (commented):
This is really great work!

I'm wondering if it also makes sense to add:
- `ExpressibleByArrayLiteral` conformance
- `public mutating func insert<S: Sequence>(contentsOf newElements: S) where S.Element == Element`

Sources/PriorityQueueModule/PriorityQueue.swift
AquaGeek: Yeah, I ran into some issues with these (though I don't recall what they were now). I'll try to remove the `Collection` initializer and re-add the `ExpressibleByArrayLiteral` conformance.

AquaGeek:
> `public mutating func insert<S: Sequence>(contentsOf newElements: S) where S.Element == Element`

I was trying to figure out if there is a performant way to do this — build a heap out of the new sequence and then merge the two?

kylemacomber:
> [ ] Figure out if `@inlinable` is required

`@inlinable` is definitely required and it dramatically improves performance!

I did a naive find and replace adding `@inlinable` to every method on `PriorityQueue`. Many operations are now 10-100x faster! It blows away `CFBinaryHeap`:

01 operations

hassila:
> > [ ] Figure out if `@inlinable` is required
> 
> 
> 
> `@inlinable` is definitely required and it dramatically improves performance!
> 
> 
> 
> I did a naive find and replace adding `@inlinable` to every method on `PriorityQueue`. Many operations are now 10-100x faster! It blows away `CFBinaryHeap`:
> 
> 

That was more like it! Awesome. Should revisit iterative vs recursive implementation again (I can do that tomorrow, have done it for a few methods with no real wins, but proper inlining may change that) - wonder how many are missing this given the comment I found googling the forums...

Sources/PriorityQueueModule/PriorityQueue.swift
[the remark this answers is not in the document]
timvermeulen: `Deque` has both `popFirst`/`popLast` and `removeFirst`/`removeLast` as well, with only the `pop_` methods returning an optional, so it looks to me like the types are now fully consistent with each other in this regard 👍

Sources/PriorityQueueModule/PriorityQueue.swift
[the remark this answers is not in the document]
timvermeulen: Private functions marked with `@inline(__always)` can definitely be inlined within the module, but it prevents the function it's inlined into from itself being inlined across module boundaries. You can easily see this by e.g. marking `popMin` with `@inlinable` — the compiler will complain that this is impossible because `_remove`, which `popMin` is implemented in terms of, is private and can therefore not be referenced from an `@inlinable` function. Hence my earlier suggestion to mark everything as `@inlinable` because the compiler will guide you to make this possible 🙂

AquaGeek:
OK, I marked everything as `@inlinable` in c33273f80c341c9c61c997dfe2daca8e88c5a5ac. Test results now are much better and in-line with what @kylemacomber got:

Benchmarks after inlining

hassila:
I did an iterative implementation as a test, seems to give a noticeable improvement when done together with some extra inlining. Created https://github.com/AquaGeek/swift-collections/pull/2 for your consideration @AquaGeek if you want it.

Recursive:
priorityqueue-recursive

Iterative:
priorityqueue-iterative

hassila:
Found out that the nice benchmark framework could do a better diff:

```
> swift-collections-benchmark results compare bench-recursive bench-iterative
Tasks with difference scores larger than 1.05:
  1.221   1.221   1.221(#76)   1.000(#0)    PriorityQueue<Int> popMin (*)
  1.201   1.201   1.201(#76)   1.000(#0)    PriorityQueue<Int> popMax (*)
  1.146   1.146   1.146(#76)   1.000(#0)    PriorityQueue<Int> insert (*)
  1.105   1.105   1.105(#76)   1.000(#0)    PriorityQueue<Int> init from range (*)

```

AquaGeek:
OK, @hassila's "iterative instead of recursive" PR has been merged. I think the only thing left here is to add invariant tests. I was planning on following the approach taken in some of the other data structures:

```swift
// PriorityQueue+Invariants.swift
extension PriorityQueue {
  #if COLLECTIONS_INTERNAL_CHECKS
  @inlinable
  @inline(never)
  internal func _checkInvariants() {
    // TODO: Check that the layers in the min-max heap are in the proper order
  }
  #else
  @inlinable
  @inline(__always)
  public func _checkInvariants() {}
  #endif  // COLLECTIONS_INTERNAL_CHECKS
}
```

Sources/PriorityQueueModule/PriorityQueue+Invariants.swift
```diff
@@ -0,0 +1,91 @@
...
+  // In the case of 7 elements total (spread across 3 levels), the checking
+  // happens in the following order:
+  //   compare >= 0: 1, 3, 4, 2, 5, 6
+  //   compare <= 1: 3, 4
+  //   compare >= 3: (no children)
+  //   compare >= 4: (no children)
+  //   compare <= 2: 5, 6
+  //   compare >= 5: (no children)
+  //   compare >= 6: (no children)
+  internal func _checkInvariants() {
```
AquaGeek, line 45: > Since heaps have some invariants that need to be maintained (min-max heaps especially), you could consider adding a method that checks these invariants just for testing purposes. With some of the operations it can be a bit tricky to keep a clear mental model of exactly what's happening.

@timvermeulen Is this what you had in mind?

AquaGeek:
OK, invariant checks have been added. Anything else we need to address here? (cc/ @kylemacomber)

kylemacomber:
> > `public mutating func insert<S: Sequence>(contentsOf newElements: S) where S.Element == Element`
> 
> I was trying to figure out if there is a performant way to do this — build a heap out of the new sequence and then merge the two?

I think it's fine to just add the naive loop that calls `insert` for every element in the sequence to start. Since we don't have to worry about ABI stability, it's trivial to substitute the implementation with a more efficient algorithm in the future.

kylemacomber:
> OK, invariant checks have been added. Anything else we need to address here? (cc/ @kylemacomber)

This is our first big addition to the collections package, so there's not really any existing precedent, but I think it'd be great to do an API review on the [collections section](https://forums.swift.org/c/related-projects/collections/72) of the forums to finalize the names.

I think something similar to the [review](https://forums.swift.org/t/api-review-filepath-syntactic-apis-version-2/44197) of the FilePath Syntactic APIs would work well. Nothing too fancy—just a dump of all the public APIs and their doc comments along with a little commentary. We should include at least couple example uses of `PriorityQueue` (either in the doc comments or the commentary)!

Sources/PriorityQueueModule/PriorityQueue+Invariants.swift
timvermeulen: Absolutely, this looks great. I was imagining we call this method from the tests somewhere but since the mutating operations already call it themselves, I don't think we need to.

Sources/PriorityQueueModule/PriorityQueue+Iterator.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    // Returns the next element in the priority queue depending on the iteration direction
+    @inlinable
+    public mutating func next() -> Element? {
+      if(_direction == .ascending){
+        return _base.popMin()
+      }
+      return _base.popMax()
+    }
```
timvermeulen, line 41: I don't expect this branch to impact the performance in any meaningful way, but if it turns out it does, we should probably split this type up into distinct types for ascending and descending iteration.

AquaGeek:
> This is our first big addition to the collections package, so there's not really any existing precedent, but I think it'd be great to do an API review on the [collections section](https://forums.swift.org/c/related-projects/collections/72) of the forums to finalize the names.
> 
> I think something similar to the [review](https://forums.swift.org/t/api-review-filepath-syntactic-apis-version-2/44197) of the FilePath Syntactic APIs would work well. Nothing too fancy—just a dump of all the public APIs and their doc comments along with a little commentary. We should include at least couple example uses of `PriorityQueue` (either in the doc comments or the commentary)!

OK, [here's the WIP doc](https://gist.github.com/AquaGeek/5865101a2fe44b312403db85143707e8) for the review. I need to fill in the details about the structure of the min-max heap and add some example usages. Once that's done I'll post it on the forum.

hassila:
Looks like a very nice start @AquaGeek - maybe append the latest performance measurements too as it will likely be one of the first questions? I can run the latest with a larger set of runs on an Intel box if you want (just let me know), would be cool to have M1 numbers too (don't have access unfortunately).

AquaGeek:
> Looks like a very nice start @AquaGeek - maybe append the latest performance measurements too as it will likely be one of the first questions? I can run the latest with a larger set of runs on an Intel box if you want (just let me know), would be cool to have M1 numbers too (don't have access unfortunately).

I don't have access to an M1, either. I'll re-run the benchmarks on my MBP and include those in the forum post (good call there!); I think that'll be fine for now. If you want to run any additional benchmarks you can post a follow-up in the thread with your results.

NevinBR:
It looks like `_minMaxHeapIsMinLevel` is always called with a `+1` at the call-site, as in:

```
_minMaxHeapIsMinLevel(index + 1)
```

Would it make sense to move the `+1` into the implementation of `_minMaxHeapIsMinLevel`, so that call-sites can just pass in `index` rather than having to remember to add one first?

hassila:
There is one thing I wonder about the performance tests - maybe a question for @lorentey if this is possible to do - looking at the output graphs from typical benchmarks, the actual construction time for the data structure is included such that we start of with a high per-item runtime (100+ns, see e.g https://forums.swift.org/t/api-review-priorityqueue-apis/49932/10) which then goes down over time for larger amount of tested items as the setup cost is amortised over the number of items.

It is interesting to know what the creation/destruction overhead is for a data structure (also as a function of "n" if reserving a certain capacity for creation and to understand if destruction is inefficient), but wouldn't it make sense to break that out into separate benchmark tests typically and keep the normal operation testing without that overhead?

It would be a more natural looking graph for most I think?

AquaGeek:
> **@NevinBR**
> Would it make sense to move the `+1` into the implementation of `_minMaxHeapIsMinLevel`, so that call-sites can just pass in `index` rather than having to remember to add one first?

Yeah, I went back and forth on this. I ended up pushing it onto the callers because the calculation is ultimately based on the `count` of items, not the index. I don't have strong feelings about it either way, though if we leave the responsibility of adding one on the callers we should probably require the parameter name in the call:

```swift
_minMaxHeapIsMinLevel(count: index + 1)

// Instead of:
_minMaxHeapIsMinLevel(index + 1)
```

CTMacUser:
In the Swift Forums discussion thread for this type, [I suggested](https://forums.swift.org/t/api-review-priorityqueue-apis/49932/34) adding an element search method.  I put it in a pull request to the repo this pull request is based on: [AquaGeek#3](https://github.com/AquaGeek/swift-collections/pull/3).

NevinBR:
> the calculation is ultimately based on the `count` of items, not the index.

That seems…needlessly pedantic to the point of causing confusion to readers.

The caller is working with an index, and wants to know “Is this index in a min-level or a max-level?”

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,537 @@
...
+
+  /// Creates an empty queue.
+  @inlinable
+  public init() {
+    storage = []
+  }
+
+  /// Inserts the given element into the queue.
+  ///
+  /// - Complexity: O(log `count`) / 2
```
WowbaggersLiquidLunch, line 67: I think the complexity is O(log `count`) / 4 instead of / 2, since other than the first comparison, we skip the parent node to compare with the grandparent.

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,537 @@
...
+      let maxIdx = storage[2] > storage[1] ? 2 : 1
+      return _remove(at: maxIdx)
+    }
+  }
+
+  /// Removes and returns the element with the lowest priority.
+  ///
+  /// The queue *must not* be empty.
+  ///
+  /// - Complexity: O(log `count`)
```
WowbaggersLiquidLunch, line 135: Should this complexity be O(log `count`) **/ 4**, since `insert`'s complexity includes the "/ 4" part?

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,537 @@
...
+}
+
+// MARK: -
+
+extension PriorityQueue {
+  /// Initializes a queue from a sequence.
+  ///
+  /// Utilizes [Floyd's linear-time heap construction algorithm](https://en.wikipedia.org/wiki/Heapsort#Floyd's_heap_construction).
+  ///
+  /// - Complexity: O(n), where `n` is the length of `elements`.
```
WowbaggersLiquidLunch, line 515: Variables (such as _n_ here) in the standard library's complexity documentation are italic. I think we should follow its style to be consistent.

timvermeulen:
> That seems…needlessly pedantic to the point of causing confusion to readers.
> 
> The caller is working with an index, and wants to know “Is this index in a min-level or a max-level?”

I couldn't agree more — it's immediately clear what it means for a node to be on a min-level or a max-level for anyone even slightly familiar with how this data structure works. "Returns `true` if `count` elements falls on a min level" is a bit of an awkward description, at least to me, because I have no intuition for what it could mean for `count` elements to fall on a certain level.

The `count > 0` precondition and the `+ 1` in every single call site are also pretty good indicators that taking an index instead would simplify the code.

Sources/PriorityQueueModule/PriorityQueue.swift
timvermeulen: We should probably stick with https://github.com/apple/swift-collections/pull/51#issuecomment-854003019 and not include any constants here and in `insert`'s docs.

Sources/PriorityQueueModule/PriorityQueue.swift
WowbaggersLiquidLunch: My mistake. / 2 is correct. Please disregard the comment above.

lorentey:
> It is interesting to know what the creation/destruction overhead is for a data structure (also as a function of "n" if reserving a certain capacity for creation and to understand if destruction is inefficient), but wouldn't it make sense to break that out into separate benchmark tests typically and keep the normal operation testing without that overhead?
> 
> It would be a more natural looking graph for most I think?

I think I would rather say that having constant overheads show up in the regular benchmark is the most useful/honest approach -- after all, separating these out would hide a very large component of performance at small sizes, which would be misleading.

In cases where a certain calculation isn't relevant, the benchmark ought to be manually written so that the calculation occurs outside of the scope of measurement. (In the case of an insertion benchmark, this can be done by reserving capacity before the measurement begins, and using `withExtendedLifetime` to also ignore deinitialization costs. This ought to be just one variant of the benchmark though -- in real world cases, allocation/deallocation costs are very much relevant, and it isn't always possible/desirable to call `reserveCapacity`.)  

In cases where the extra calculation cannot be separated from the thing we're measuring, then why exactly would we want to hide it? Artificially massaging benchmark curves to better match the expected asymptotic behavior at small sizes wouldn't be useful. Most real-world charts that plot average per-element time naturally have a lopsided U shape -- getting rid of the initial downward slope _by tweaking the benchmarking methodology_ would be a non-goal, I think.

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,537 @@
...
+
+    _checkInvariants()
+  }
+
+  /// Inserts the elements in the given sequence into the priority queue.
+  ///
+  /// - Parameter newElements: The new elements to insert into the queue.
+  ///
+  /// - Complexity: O(n * log `count`), where `n` is the length of `newElements`.
+  public mutating func insert<S: Sequence>(contentsOf newElements: S) where S.Element == Element {
```
AquaGeek, line 532: Adding `@inlinable` here results in ~10x speedup.

AquaGeek:
OK, I've integrated most of the feedback and split `MinMaxHeap<Element: Comparable>` from `PriorityQueue<Value, Priority: Comparable>`. My changes are on a [separate branch](https://github.com/AquaGeek/swift-collections/compare/priority-queue...AquaGeek:pq-rework) if anyone wants to have a look.

@kylemacomber and @lorentey: How do you want to proceed from here?
* Should I merge those changes into this same `priority-queue` branch and continue the review in this PR?
* How do you want to handle @CTMacUser's PR (AquaGeek/swift-collections#3)? Do you want that to be part of this chunk of work, or do you want to review/merge it separately?
* Do you want `replaceMin` and `replaceMax` to be added as part of this initial chunk of work?
* How do you want to handle the significant changes involved in splitting the two types apart again in regards to the API review in the forum? Should that topic be closed as "returned for review" and a new one opened once we've reached a good point on the code side with the full API docs again of both types, or should I update that thread with the updated docs for the two types?

Thanks in advance!

AquaGeek:
Here are preliminary benchmarks of `PriorityQueue<Value, Priority>` vs `MinMaxHeap<Element>` vs `CFBinaryHeap`:

## Insertions
pq-insert

## Removals
pq-remove

kylemacomber:
I'm open to other directions, but I think it might be good to rename this type to `Heap` and land it with no additional features (maybe into a PriorityQueue feature branch). Then we can open up other PRs for the `PriorityQueue` work and some of the other enhancements people have proposed.

Also it might be worth including the API review document as part of this initial commit.

hassila:
> I think I would rather say that having constant overheads show up in the regular benchmark is the most useful/honest approach -- after all, separating these out would hide a very large component of performance at small sizes, which would be misleading.
> ...
> 
> In cases where the extra calculation cannot be separated from the thing we're measuring, then why exactly would we want to hide it? Artificially massaging benchmark curves to better match the expected asymptotic behavior at small sizes wouldn't be useful. Most real-world charts that plot average per-element time naturally have a lopsided U shape -- getting rid of the initial downward slope _by tweaking the benchmarking methodology_ would be a non-goal, I think.

Thanks for the reply - the reason for "hiding it" would be that many performance sensitive implementations recycle data structures to remove the init/deinit overhead - in this particular case, we'd probably have a long life cycle for the priority queues, but constantly add/remove a smallish number of items (perhaps 10-10.000 depending on load), so the performance characteristics for long-lived data structures where you have a smallish amount of data items are effectively opaque as the init/deinit overhead overshadows the actual run time for the operations. 

But I'm not really arguing for hiding it completely, but rather suggest that it might be interesting to understand init/deinit costs for the data structure _separately_ from the actual operation -  I was just thinking it might make sense to split that out to a separate test (init/deinit times for data structure with N elements), fundamentally splitting the current single curve into three curves, one for init, one for deinit, one for the actual real processing (which would be relevant for long-lived structures).  That would allow users with long-living data structures to understand performance characteristics as well as users who have transient data structures (who would need to look at two or all three of the curves in that case).

Anyway, I'll be ok with the current setup too, one gets a worst-case performance estimate when looking at the lowest point of the curve - thanks for explaining your reasoning.

timvermeulen:
> * Should `replaceMin` and `replaceMax` to be added as part of this initial chunk of work? (Again, I think I'd prefer that those be follow-up PRs.)

@AquaGeek For what it's worth, I've written a draft implementation for those [here](https://github.com/AquaGeek/swift-collections/compare/priority-queue...timvermeulen:priority-queue-replace). But I agree about landing the data structure first, also because there are some unresolved questions that we'd need to answer 🙂

AquaGeek:
@kylemacomber I opened a new PR (#61) with just the `Heap` (and all the commits leading to that). I figure we can retarget that to a new `priority-queue` branch if needed. Do we want to do another separate API review for that in the forums, or is the feedback we received from the `PriorityQueue` API review (and the changes to split the two types apart and loosen the `Comparable` requirement on the `PriorityQueue` element type) sufficient?

AquaGeek:
OK, I've force-pushed the commits to this branch that add the priority queue implementation on top of `Heap` (which @lorentey merged yesterday 🎉). I'll need to take a pass and update the docs (and there are probably a few other things I need to tweak), but it should be largely ready for re-review.

Sources/PriorityQueueModule/PriorityQueue+Iterator.swift
AquaGeek: This has been done (for both `PriorityQueue` and `Heap`).

Benchmarks/Benchmarks/Library.json
```diff
@@ -791,15 +797,21 @@
           "title": "insert",
           "tasks": [
             "Heap<Int> insert",
-            "Heap<Int> insert(contentsOf:)"
+            "Heap<Int> insert(contentsOf:)",
+            "PriorityQueue<Int, Int> insert",
+            "PriorityQueue<Task, Int> insert"
```
AquaGeek, line 802: @lorentey Should we pull the priority queue stuff out into a separate library JSON? (Or add a separate priority queue library JSON in addition to adding these benchmarks here?)

lorentey:
@AquaGeek In hindsight, I think the single huge library file isn't very useful -- I find I almost never want to generate benchmarks for every single data structure in this package (it's just too slow), but rather I concentrate on one collection at a time.

I started to split recent library additions into separate files -- for example, we have a smaller library specifically for `Heap` in the [Benchmarks/Libraries/Heap.json](https://github.com/apple/swift-collections/blob/main/Benchmarks/Libraries/Heap.json) file. It would make sense to add `PriorityQueue` benchmarks to that file, too. (I think it is okay not to update the primary library file -- I expect I'll eventually replace its contents with an inclusion mechanism, or just remove it altogether.)

I believe the `PriorityQueue` charts would work better as new additions, separate from the existing `Heap` ones. (I find a chart with more than 5-6 tasks is very hard to follow.) In addition to standalone charts that only show `PriorityQueue` tasks, we can also have a set of charts for `Heap` vs `PriorityQueue` comparisons for each (family of) operations. (It may be useful to make the `Element` type of `Heap` a custom struct combining the keys/values in `PriorityQueue` to make these fair. It's extra work though, and I don't think it's critical -- comparing against `Heap<Int>` would still be interesting. I don't think we need to directly compare `PriorityQueue` to C++ constructs, but that's an option too.)

AquaGeek:
@lorentey

> @AquaGeek In hindsight, I think the single huge library file isn't very useful -- I find I almost never want to generate benchmarks for every single data structure in this package (it's just too slow), but rather I concentrate on one collection at a time.

Yeah. While I was going through this, I didn't fully understand how libraries work, so I was commenting out the irrelevant benchmarks in `main.swift`. 😆

> I started to split recent library additions into separate files -- for example, we have a smaller library specifically for `Heap` in the [Benchmarks/Libraries/Heap.json](https://github.com/apple/swift-collections/blob/main/Benchmarks/Libraries/Heap.json) file. It would make sense to add `PriorityQueue` benchmarks to that file, too. (I think it is okay not to update the primary library file -- I expect I'll eventually replace its contents with an inclusion mechanism, or just remove it altogether.)

I originally had the benchmarks in the primary library file as well, but I removed them from that and created a new separate file for them (`Benchmarks/Libraries/PriorityQueue.json`).

> I believe the `PriorityQueue` charts would work better as new additions, separate from the existing `Heap` ones. (I find a chart with more than 5-6 tasks is very hard to follow.) In addition to standalone charts that only show `PriorityQueue` tasks, we can also have a set of charts for `Heap` vs `PriorityQueue` comparisons for each (family of) operations. (It may be useful to make the `Element` type of `Heap` a custom struct combining the keys/values in `PriorityQueue` to make these fair. It's extra work though, and I don't think it's critical -- comparing against `Heap<Int>` would still be interesting. I don't think we need to directly compare `PriorityQueue` to C++ constructs, but that's an option too.)

I updated the `Heap` benchmarks to also include ones for a small `Task` struct. Here are the comparative benchmarks (MacBook Pro (16-inch, 2019), 2.3 GHz 8-Core Intel Core i9):

01 insert
02 pop

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,164 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Swift
+
+/// A double-ended priority queue built on top of `Heap`.
+public struct PriorityQueue<Value, Priority: Comparable> {
+  public typealias Element = (value: Value, priority: Priority)
+
+  public struct Pair: Comparable {
```
AquaGeek, line 18: I'm not completely sold on this name. Anyone have a better one? `Element`?

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,168 @@
...
+
+/// A double-ended priority queue built on top of `Heap`.
+public struct PriorityQueue<Value, Priority: Comparable> {
+  public typealias Pair = (value: Value, priority: Priority)
+
+  @usableFromInline
+  struct _Element: Comparable {
+    @usableFromInline let value: Value
+    let priority: Priority
+    let insertionCounter: UInt64
```
AquaGeek, line 22: Anyone have any thoughts on this approach? Is using `UInt64` overkill for this? Should we overflow back to 0 on insertion instead of trapping?

jnssnmrcs:
What about a decreasePriority/increasePriority functions? It's used by some algorithms, for example Dijkstra's or A*, to become faster.

datwelk:
What is the status of this PR, is it pending review still? @AquaGeek

AquaGeek:
> What is the status of this PR, is it pending review still? @AquaGeek

Yes, it's still pending review. In particular, I'm looking for feedback on the best way to handle FIFO ordering — right now, I'm using a `UInt64` insertion counter.

I'd love to get this landed.

datwelk:
@AquaGeek Have you considered using a timestamp such as Date() to enforce FIFO in case of equal priorities instead of the counter? What would be the disadvantages of that approach?

AquaGeek:
> @AquaGeek Have you considered using a timestamp such as Date() to enforce FIFO in case of equal priorities instead of the counter? What would be the disadvantages of that approach? I guess precision.

I think that would be effectively the same thing but with the downside of depending on `Foundation`. There may be performance or space differences worth digging into.

hassila:
> > @AquaGeek Have you considered using a timestamp such as Date() to enforce FIFO in case of equal priorities instead of the counter? What would be the disadvantages of that approach? I guess precision.
> 
> I think that would be effectively the same thing but with the downside of depending on `Foundation`. There may be performance or space differences worth digging into.

If exploring this, I'd aim for using `Instant` then from 5.7 to keep it Foundation-free (https://github.com/apple/swift-evolution/blob/main/proposals/0329-clock-instant-duration.md). Space-wise it would be twice (128 bits), performance of getting the current time is fairly decent these days (used to suck big time), but hardly as cheap as maintaining a counter - which is the best thing I can think of at least for maintaining FIFO.

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,168 @@
...
+  public mutating func popMin() -> Value? {
+    _base.popMin()?.value
+  }
+
+  /// Removes and returns the item with the highest priority, if available.
+  ///
+  /// - Complexity: O(log `count`)
+  @inlinable
+  public mutating func popMax() -> Value? {
+    _base.popMax()?.value
```
e1t0n, line 123: Can we also return the priority along with value like in Python?
```
next = q.get()
print(next)
//(1, 'Jones')
```

lorentey:
My gut feeling is that if someone wants to ensure that this preserves insertion order for items with the same priority, then they can simply define their Priority type to account for this.

Using `Date` (or, rather, `ContinuousClock.now`) would not be appropriate -- clock do not return strictly monotonically increasing instants.

lorentey:
Then again, if they're defining a type anyway, they could just as well define a custom `Comparable` wrapper over their element type, and just use `Heap`.

I don't quite get the value this adds over `Heap` at all. The case for `PriorityQueue` isn't nearly as self-evident as it is with `Set` vs `Dictionary` -- `Dictionary` adds a wealth of functionality that isn't directly expressible using `Set`. `PriorityQueue` is effectively just a glorified typealias:

```swift
public struct Prioritized<Element, Priority: Comparable>: Comparable {
  let priority: Priority
  let element: Element

  init(_ element: Element, _ priority: Priority) { 
    self.priority = priority
    self.element = element
  }

  static func ==(left: Self, right: Self) -> Bool { left.priority == right.priority }
  static func <(left: Self, right: Self) -> Bool { left.priority < right.priority }
}

public typealias PriorityQueue<Element, Priority: Comparable> = 
  Heap<Prioritized<Element, Priority>>
```

Can y'all remind me of the point in defining a separate `PriorityQueue` type?

AquaGeek:
> Can y'all remind me of the point in defining a separate PriorityQueue type? Is it just for convenience?

Honestly, it has been so long that I've lost a lot of the context here.

> If stable ordering is needed often enough to deserve its own place here, then ideally it should be offered for Heap as well, by a StableHeap wrapper type.

At the moment, that would have been my primary argument for `PriorityQueue` — I imagine the typical use case would prefer FIFO ordering and not having to conform to `Comparable` (i.e. splitting out the element from the priority).

lorentey:
Not having to conform to `Comparable` is not _necessarily_ a primary goal here, but I certainly see how guaranteeing stable ordering would be annoying to implement, especially if one had to repeatedly do that.

My hesitation is really just me getting a bit worried about the hidden costs of having to store a full UInt64 value with every single item, just on the off chance some of them might compare equal. Clients that do not care about preserving insertion order shouldn't have to pay the storage overhead of these. (A hidden 8-byte value usually doesn't sound too bad, but it can end up significant enough to be worth worrying about.)

Automatically & implicitly bundling stable ordering with the Element+Priority separation sounds unappealing to me -- both aspects are interesting ideas on their own, but combining them together under the name `PriorityQueue` seems less palatable on the whole.

What if we waited until we had some actual users for this module before we start elaborating it? If potential adopters came back complaining about the pain of having to define wrapper types and/or asking how they could preserve insertion order, then we'd probably be in a better position to figure out the right way to serve them. (Which may well turn out to be your original design -- I may just not be seeing that far ahead!)

Note: I'm not aware of any high-profile heap implementations in other languages that provide out-of-the-box support for preserving insertion order. Then again, I only did a quick & very basic search and may have missed something obvious.

(Now of course we could also preserve insertion ordering using a different design -- such as using UInt32 values, and renumbering all existing items if we ever wrap around, or by tweaking the underlying Heap implementation in some way. But then again, the effort of thinking about & implementing these may well prove unnecessary, and why bother until we know we'll need it...)

AquaGeek:
> Not having to conform to `Comparable` is not _necessarily_ a primary goal here, but I certainly see how guaranteeing stable ordering would be annoying to implement, especially if one had to repeatedly do that.
> 
> My hesitation is really just me getting a bit worried about the hidden costs of having to store a full UInt64 value with every single item, just on the off chance some of them might compare equal. Clients that do not care about preserving insertion order shouldn't have to pay the storage overhead of these. (A hidden 8-byte value usually doesn't sound too bad, but it can end up significant enough to be worth worrying about.)

Yeah, I think that's valid. I'm hesitant to commit to too many design decisions in absence of compelling use cases.

There's always the "if you care about eking out every last drop of performance you can drop down to `Heap` directly" argument. 🤷

> What if we waited until we had some actual users for this module before we start elaborating it? If potential adopters came back complaining about the pain of having to define wrapper types and/or asking how they could preserve insertion order, then we'd probably be in a better position to figure out the right way to serve them. (Which may well turn out to be your original design -- I may just not be seeing that far ahead!)

@phausler was one of the people who brought up FIFO ordering in [the forum thread](https://forums.swift.org/t/api-review-priorityqueue-apis/49932/27). The comparison was against `NSOperationQueue`.

The original use case for me starting to implement this at all was a priority queue for pending network requests. FIFO ordering is important, as we don't want a request that was enqueued first waiting forever in the queue because subsequent requests with the same priority happen to get inserted into a slot ahead of it in the heap. That being said, we haven't cut over to this library's implementation yet, partly because we were hoping to get the FIFO ordering and the removal of the `Comparable` requirement provided by `PriorityQueue` first (network requests being `Comparable` directly doesn't really make sense, though I guess we could always create a wrapper type).

> (Now of course we could also preserve insertion ordering using a different design -- such as using UInt32 values, and renumbering all existing items if we ever wrap around, or by tweaking the underlying Heap implementation in some way. But then again, the effort of thinking about & implementing these may well prove unnecessary, and why bother until we know we'll need it...)

I mainly used `UInt64` out of an abundance of caution. `UInt32` still provides 4B inserts before rolling over — my first reaction is "nobody will hit that," but network servers that run for a long time might. There is probably prior art to draw on here. One quick thought I had was to zero the insertion counter when the queue becomes empty.

Sources/PriorityQueueModule/PriorityQueue.swift
Cay-Zhang: I'm also confused about why we can't get back the priorities we previously set.

Sources/PriorityQueueModule/PriorityQueue.swift
```diff
@@ -0,0 +1,168 @@
...
+    @usableFromInline
+    static func == (lhs: Self, rhs: Self) -> Bool {
+      lhs.priority == rhs.priority
+    }
+
+    @usableFromInline
+    static func < (lhs: Self, rhs: Self) -> Bool {
+      if lhs.priority < rhs.priority {
+        return true
+      } else if lhs.priority == rhs.priority {
```
AquaGeek, line 40: One issue we've run into in our upstream implementation of this, is that this `Comparable` implementation only works when trying to dequeue the minimum element. Dequeueing the maximum element doesn't respect FIFO ordering due to the default implementation of `>` being to flip the arguments and use `<` (i.e. `a > b` ⇢ `b < a`). Implementing `>` to also account for the insertion counter starts violating invariants in the heap — it creates weird situations where an element can be both greater than and less than another if it was inserted before and has the same priority.

One way we've come up with to fix this is to only insert the priorities into the underlying heap and use them as a key for looking up an array (or linked list or deque) in a dictionary that actually contains the values. That eliminates the need to store a `UInt64` per element but comes at some cost (especially if you use an array) and requires that the `Priority` be `Hashable`. None of these changes affect the underlying `Heap`.

```swift
struct PriorityQueue<Value, Priority: Hashable> {
    var _base: Heap<Priority>
    var _elements: Dictionary<Priority, Deque<Value>>

    ...
}
```

Are there other approaches for maintaining FIFO ordering?

lorentey:
> One issue we've run into in our upstream implementation of this, is that this Comparable implementation only works when trying to dequeue the minimum element. Dequeueing the maximum element doesn't respect FIFO ordering due to the default implementation of > being to flip the arguments and use < (i.e. a > b ⇢ b < a). 

This is a good point, but only as far it underscores a point. From my viewpoint, the fact that `popMin`'s ordering is precisely the reverse of `popMax` is not a bug; rather, it is the entire point.

Expecting `popMin` and `popMax` to both observe FIFO ordering would be an undesirable semantic complication that I don't believe would carry its weight -- for the data structure to make sense, the `<` relation must define a total ordering, and the pop operations need to observe it.

Clients have full flexibility resolve the ordering of items as they wish, by conforming their `Element` (or if you like, `Priority`) type to `Comparable` as they choose. If a client wishes to have `popMin` return elements with equal priority in FIFO order, then it's straightforward to implement that by adding a serial number to the custom element type, then defining `<` to compare that in addition to the priority value. `popMax` will then return items in the opposite order of that, by definition. If a client wants `popMax` to use FIFO ordering, then they can simply reverse the polarity of the serial number comparison.

I believe `Heap` is a great implementation (given the constraints of today's Swift), and it is ready to ship as is. If we also want to add a second heap-like type, I'd need that to provide something that (1) is useful to many clients, but (2) isn't trivial to implement. 

The idea of adding a serial number to incoming items to implement a FIFO `popMin` operation (like this PR does) would pass the first criterium, but it doesn't seem complex enough to be worth the expense of adding it to this package. (Keep in mind that this package has never seen a feature release beyond 1.0, and it has a very significant unreleased backlog of highly important work -- potential new additions need to pass an incredibly high level of importance check.)

Adding a second heap-like data structure that provides both a FIFO `popMin` **and** a FIFO `popMax` (as suggested) would certainly pass the second criterium, but I do not see how can it possibly pass the first. The double-ended nature of our min-max heap is a bonus convenience feature that we got for free; I'm not aware of important use cases that inherently require it, much less ones that need both directions to implement a FIFO ordering.

Given how backlogged this package is, I don't see us shipping second-order additions like this in the foreseeable future -- accordingly, I'm going to close this PR. Apologies for leaving it open this long!

I'm not against trying again after we've managed to ship the current set of data structures in limbo -- however, my concerns above will likely still apply. An example of a perhaps more interesting direction (that could better satisfy both criteria above) might be to have a `PriorityQueue` type that supports an O(log(n)) `remove` operation.

fbartho:
Sorry for resurrecting the discussion here if people are exhausted about it; but if this library is never going to add PriorityQueue -- would an acceptable alternative be some Examples in the docs on how to use the Heap type to roll your own PriorityQueues using Heap?

Looking at the diff in this PR, there's a ton of details that could be omitted if users are implementing the PriorityQueue they need, so I think a short-enough example would be small enough to fit in the docs?

PR #78, opened 2021-08-07
[Heap] Performance tweaks

### LINKED ISSUE #69: [Heap] Improved consistency checking
The current `Heap._checkInvariants` method doesn't fully validate the min-max invariants -- it just checks to see if the property holds for the immediate children of any node.

Instead, we should compare the value stored in every node with a currently expected range of values.

E.g., in the tree below, when checking descendants of 11, each value needs to be within the range 11 ... 41. Each new level narrows this range a bit further (by refining the lower or upper bounds, depending on the polarity of the node).

```
 level 0:                 8
 level 1:         71              41
 level 2:     31      10      11      16
 level 3:   46  51  31  21  13
```

Implementing this would mean that we'd have to keep track of a stack of expected ranges in addition to their associated indices. (Or switch to a recursive implementation.)

### LINKED ISSUE #71: [Heap] Switch to a custom _Index type that incrementally tracks depth
The `Heap` implementation currently uses naked integers as indices, and uses `_binaryLogarithm` to determine whether a particular index addresses a min or max node. This is a relatively expensive(*) operation that may have some measurable performance impact.

(*) Well, at least compared to a mov?

Instead, try switching to a custom _Index struct containing both a storage position and its associated depth.

Beyond a potential performance win, this would also allow for better code organization -- by allowing us to move the utility methods for retrieving the parents/children of an index to the Index type itself:

```swift
struct _Index: Comparable {
  let offset: Int
  let level: Int

  func parent() -> _Index?
  func grandParent() -> Index?
  func leftChild() -> _Index
  func rightChild() -> _Index
}
```

### LINKED ISSUE #75: [Heap] Express heap operations on UnsafeMutableBufferPointer
`Heap` currently uses regular array subscripting to access the contents of its storage. Array access is pretty fast, but it includes index validation that is (hopefully) unnecessary in `Heap`'s case. I expect the optimizer is able to get rid of these index validations in some (but not all) cases; but it probably makes sense to not have them in the first place.

Rework the low-level heap implementation to operate on UnsafeMutableBuffer storage instead, turning index checks into debug-mode assertions rather than unconditional preconditions.

To do this, it would make sense to introduce a `Heap._UnsafeHandle` or `_UnsafeHeap` struct that consists of an UnsafeMutableBufferPointer value, and move most heap operation to that. We'll need to have closure-based `read` and `update` helpers that expose a read-only (or read-write) view of the Array storage. (Similar to how Deque and OrderedSet's hash tables work.)

lorentey:
- `@inline(__always)` has a detrimental effect when used on large functions. Let's not do that.
- Speed up invariant checking with COLLECTIONS_INTERNAL_CHECKS (resolves #69)
- Replace naked ints with `_Node`, precalculating levels for each offset (resolves #71)
- Rebase heap operations on UMBP instead Array, eliminating unnecessary index validation (resolves #75)
- ~Add some (slightly questionable) effects attributes to heap internals.~
- Optimize removals and heap construction

### Checklist

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/PriorityQueueModule.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/PriorityQueueModule.xcscheme
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      codeCoverageEnabled = "YES">
       <Testables>
          <TestableReference
             skipped = "NO">
--- a/Sources/PriorityQueueModule/Heap+Invariants.swift
+++ b/Sources/PriorityQueueModule/Heap+Invariants.swift
@@ -14,78 +14,39 @@ import Foundation
 
 extension Heap {
   #if COLLECTIONS_INTERNAL_CHECKS
-  /// Iterates through all the levels in the heap, ensuring that items in min
-  /// levels are smaller than all their descendants and items in max levels
-  /// are larger than all their descendants.
-  ///
-  /// The min-max heap indices are structured like this:
-  ///
-  /// ```
-  /// min               0
-  /// max        1             2
-  /// min      3   4         5     6
-  /// max     7 8 9 10     11 12 13 14
-  /// min    15...
-  /// ...
-  /// ```
-  ///
-  /// The iteration happens in depth-first order, so the descendants of the
-  /// element at index 0 are checked to ensure they are >= that element. This is
-  /// repeated for each child of the element at index 0 (inverting the
-  /// comparison at each level).
-  ///
-  /// In the case of 7 elements total (spread across 3 levels), the checking
-  /// happens in the following order:
-  ///
-  /// ```
-  /// compare >= 0: 1, 3, 4, 2, 5, 6
-  /// compare <= 1: 3, 4
-  /// compare >= 3: (no children)
-  /// compare >= 4: (no children)
-  /// compare <= 2: 5, 6
-  /// compare >= 5: (no children)
-  /// compare >= 6: (no children)
-  /// ```
+  /// Visits each item in the heap in depth-first order, verifying that the
+  /// contents satisfy the min-max heap property.
   @inlinable
   @inline(never)
   internal func _checkInvariants() {
     guard count > 1 else { return }
-    var indicesToVisit: [Int] = [0]
+    _checkInvariants(index: 0, min: nil, max: nil)
+  }
 
-    while let elementIdx = indicesToVisit.popLast() {
-      let element = _storage[elementIdx]
-
-      let isMinLevel = _minMaxHeapIsMinLevel(elementIdx)
-
-      var descendantIndicesToVisit = [Int]()
-
-      // Add the children of this element to the outer loop (as we want to check
-      // that they are >= or <= their descendants as well)
-      if let rightIdx = _rightChildIndex(of: elementIdx) {
-        descendantIndicesToVisit.append(rightIdx)
-        indicesToVisit.append(rightIdx)
+  @inlinable
+  internal func _checkInvariants(index: Int, min: Element?, max: Element?) {
+    let value = _storage[index]
+    if let min = min {
+      precondition(value >= min,
+                   "Element '\(value)' at index \(index) should be >= '\(min)'")
+    }
+    if let max = max {
+      precondition(value <= max,
+                   "Element '\(value)' at index \(index) should be <= '\(max)'")
+    }
+    if _minMaxHeapIsMinLevel(index) {
+      if let left = _leftChildIndex(of: index) {
+        _checkInvariants(index: left, min: value, max: max)
       }
-      if let leftIdx = _leftChildIndex(of: elementIdx) {
-        descendantIndicesToVisit.append(leftIdx)
-        indicesToVisit.append(leftIdx)
+      if let right = _rightChildIndex(of: index) {
+        _checkInvariants(index: right, min: value, max: max)
       }
-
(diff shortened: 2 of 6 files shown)


lorentey:
@swift-ci test

AquaGeek:
I'd be interested to see the performance difference here.

lorentey:
Removing `@inline(__always)` [turned out to be a significant regression](https://github.com/apple/swift-collections/files/6950041/diff.pdf). 😅This won't land as is.

I'd like to keep an eye on code size though -- always inlining large(ish) functions is generally a bad move. We'll see if followup work can make some of this work better.

lorentey:
The thing I'm worried about is that the sequence-based initializer, insertions, removals are currently all emitted as large monolithic functions, with no code sharing between them. 

They aren't _terribly_ large (e.g., the sequence initializer is 1200 bytes on x86_64), but they get duplicated with each specialization, and this can add up.

If we can arrange things so that the bubbleUp/Down function calls remain intact without ruining performance, then that would calm my nerves a bit. :-)

lorentey:
@swift-ci test

lorentey:
Benchmark progression:

- cc690696: Original
- 891eb7a7: `_Node`
- 6822d83b: `_UnsafeHandle`
- 454da8fb: `@inline(__always)` → `@_effects(releasenone)`

chart
chart
chart
chart
chart
chart

lorentey:
@swift-ci test

lorentey:
This gets us to ~parity with `std::priority_queue`, even slightly beating it on large heaps. We still have a slight disadvantage at small sizes, though. (https://github.com/apple/swift/pull/38867 might help close that gap.)

01 initializers
02 insert
03 pop

lorentey:
Cc @AquaGeek 

Most of these are dirty tricks that reduce code readability (with the possible exception of `_Node`), but they are probably worth it for the performance boost.

The addition of the `@_effects` attributes isn't entirely justified; if we have to take those out, then that will impact `insert` quite a bit. 🤔

hassila:
Looks like interesting progress - just one thought, wouldn’t it make sense to use e.g https://github.com/skarupke/heap as the reference for the heap performance rather than the std-prioq?

xwu:
Seems `891eb7a`: `_Node` is a straight-up win in performance over the original, and the rest speed up some and regress some here and there.

lorentey:
> just one thought, wouldn’t it make sense to use e.g https://github.com/skarupke/heap as the reference for the heap performance rather than the std-prioq? (And compare the actual PriorityQueue with the std-prioq).

`std::priority_queue` implements a heap, closely matching our `Heap` (although one-sided) -- I think the comparison against it is fair (although these are never really apple-to-apples comparisons). The `PriorityQueue` concept doesn't directly exist in the STL, although (depending on the internal impl) I expect `std::priority_queue` with a custom comparator would still serve as a useful baseline for it.

`std::priority_queue` being one-sided does not make it a bad baseline -- it solves a simpler problem than `Heap`, but our hypothesis was that min-max heaps will provide matching or better performance anyway.

It would be interesting to see benchmarks against @skarupke's implementation! I'll see if I have time to run some experiments over the weekend. 🤔 We won't add those benchmarks to this repository though. 

Another interesting source of data structure implementations would be LLVM's ADT library.

I don't think we need to add an extensive library of reference benchmarks though -- verifying against some plausible baseline is useful during the initial implementation effort, but afterwards I believe we should be primarily comparing potential changes against earlier versions of this package, not against other languages.

lorentey:
One thing that we still need to address is that the code does not currently implement the optimization techniques that make https://github.com/skarupke/heap fast. We need to reorder comparisons in the grandchild min/max search, and use a more clever heapify loop.

lorentey:
The results above were collected on a 5-years old MacBook Pro. M1 results below!

The `insert` benchmark results are significantly different on M1 -- for insertions on M1, 6822d83 was a significant regression, not an improvement. This may be related to apple/swift#38867, although I expected swift_retain/relase would be less costly on the M1, not more. More analysis is needed.

chart
chart
chart
chart
chart
chart

Versus baselines:

01 initializers
02 insert
03 pop

karwa:
Drive-by comment:

Have you tried using a `ContiguousArray` for `Heap._storage`? It may help reduce code size, as regular `Array` needs to account for potential bridged storage. There may be places where the compiler can’t prove it _isn’t_ bridged.

If you need it as an `Array` elsewhere, constructing an `Array` from a `ContiguousArray` is basically free. It just retains the existing storage, it doesn’t copy.

lorentey:
Oh yeah, that's a good suggestion; OrderedSet does the same. It shouldn't* have an effect on the current benchmarks, as `Element` is always Int and `Array<Int>` is statically known to never be verbatim-bridged -- but I expect there would be a benefit for class-keyed heaps.

\* Although ContinuousArray sometimes misses some of the optimization love that Array gets, so it could be slower sometimes -- for instance, I forgot to apply apple/swift#38867 to ContiguousArray just now. 🤦

lorentey:
c4282ca7b9ed46a9c4dfcf02e60f41210aea2463 improves popMin/popMax (and the sequence initializer) by reviewing trickleDown and optimizing things:

- Replace swapAt with a scheme where we keep a hold in the storage buffer
- Slightly shorten min/max dependency chain in primary sink loop

In exchange, we get even less readable code.

M1 benchmarks:

chart
chart
chart
chart
chart
chart

lorentey:
@swift-ci test

lorentey:
OK, I'm happy with where we are now. I think `insert` could probably be improved if we could maintain a hole in the storage array rather than swapping things in it -- but that would need an append variant of `Array.init(uninitializedCapacity:initializingWith:)` so that's for later.

Final measurements with a toolchain that includes https://github.com/apple/swift/pull/38867 & https://github.com/apple/swift/pull/38898: (taken on a Mac Pro this time)

01 Heap Int  init from range
02 Heap Int  init from buffer
03 Heap Int  popMin
04 Heap Int  popMax
05 Heap Int  insert(contentsOf:)
06 Heap Int  insert

lorentey:
@swift-ci test

compnerd:
This seems to have regressed the Windows build: https://dev.azure.com/compnerd/3133d6ab-80a8-4996-ac4f-03df25cd3224/_apis/build/builds/53542/logs/239

PR #83, opened 2021-08-15
BitArray and BitSet data structures

MahanazAtiqullah:
The bit array data structure stores boolean values by reresenting them as the bits of an unsigned integer. This allows for a memory optimization and potentially a performance optimization for operations such as bitwise operations. The API works as though one is operating on an array of booleans. The bit set data structure utilizes the bit array as its storage, but has the API of a set, where it stores the indices of the true values.

### Checklist


Sources/BitArrayModule/BitSet/BitSetOperations.swift
```diff
@@ -0,0 +1,24 @@
...
+  
+  public func contains(_ member: Int) -> Bool {
+    
+    precondition(member >= 0, "Input must be a positive number")
+    if (member >= storage.count) {"[0, 3, 4, 6]"
+      return false
+    }
+    var set = Set<Int>()
+    set.union(set)
+    return storage[member]
```
NevinBR, line 21: What’s going on here? That `"[0, 3, 4, 6]"` seems out of place, and the `var set = Set<Int>()` part appears to be unnecessary.

Sources/BitArrayModule/BitSet/BitSetSetAlgebra.swift
```diff
@@ -0,0 +1,123 @@
...
+    } else {
+      for i in 0..<other.storage.storage.count {
+        self.storage.storage[i] ^= other.storage.storage[i]
+      }
+    }
+  }
+  
+  public static func == (lhs: BitSet, rhs: BitSet) -> Bool {
+    return (lhs.storage == rhs.storage)
+  }
```
NevinBR, line 121: It seems a bit odd to have an extension granting `Equatable` conformance, which contains so many other things besides just an implementation of the `==` operator.

Also, this implementation appears to be at best unintuitive, and perhaps outright incorrect, considering how element removal currently works:

```
let a: BitSet = []
var b = a
b.insert(100)
b.remove(100)
print(a == b)           // I would expect `true`, but I believe it is currently `false`
```

Sources/BitArrayModule/BitSet/BitSetOperations.swift
MahanazAtiqullah: Yes, they do not belong. They are tinkerings added later and not part of the implementation. They were merged accidentally and will be removed later.

Sources/BitArrayModule/BitSet/BitSetSetAlgebra.swift
MahanazAtiqullah: Yes, you pointed that out earlier in a comment. It will be corrected soon and I will answer that comment soon along with other comments

lorentey:
Merging to feature branch `feature/BitSet`.

lorentey:
@scott-d-bowen We do not monitor comments on closed PRs. If you're missing some functionality, please open a new issue to describe what you need.

PR #86, opened 2021-08-19
Stop depending on swift-collections-benchmark

lorentey:
Instead, move the benchmark suite into its own private nested package, under `Benchmarks/`.

The recommended way to build & run benchmarks is via the `Utils/run-benchmarks.sh` script, which continues to work as before with this change.

### Checklist


lorentey:
@swift-ci test

lorentey:
This will also need to be cherry-picked on the release/1.0 branch.

tomerd:
looks great.

is checking-in the resolved file intentional?

tomerd:
cc @neonichu @abertelrud

lorentey:
> is checking-in the resolved file intentional?

It is! I'd prefer if we ran the benchmarks using a consistent codebase, in as much as that's possible. (Not that I'd expect picking up a new argument parser would directly affect measurements -- but subtle function alignment changes within the CollectionsBenchmark module (such as when picking up a new version of a dependency) can measurably perturb benchmarks. The runner script tries to align modules on page boundaries, but this doesn't help if a function gets slightly offset within the module.)

(If this proves unnecessary, and/or having to manually update dependencies becomes too much of a chore, then we can always remove the lock later.)

lorentey:
There is no point in having the top-level resolved file though; I'll remove it.

lorentey:
@swift-ci test

abertelrud (approved):
Thanks!  This is much nicer for clients, I think.

PR #114, opened 2021-10-03
Better `Sequence` support for `Heap`

CTMacUser:
This request expands on `Sequence` support for `Heap`s.

- To match the ordered views, the property for the unordered view has a type-alias.
- The ordered views fully implement `underestimatedCount`.
- The ordered views fully implement element search.  This is due from...
- `Heap` now has an element-search method.

That last method has two implementations inside, breadth- and depth-first search.  Those need to be benchmarked against each other.  They also need to be compared against copying an unordered view and running a (linear) search on that.  We need various sizes and search target locations (before all actual values, after them, between them, an actual match).  The complexity level is a guesstimate; it could be worst-case linear, at least for breath-first.  But those are just the time complexities, the scratch space may be significant.

### Checklist

### DIFF
--- a/Sources/PriorityQueueModule/Heap+OrderedViews.swift
+++ b/Sources/PriorityQueueModule/Heap+OrderedViews.swift
@@ -32,6 +32,11 @@ extension Heap {
     public var underestimatedCount: Int {
       _base.count
     }
+
+    @inlinable @inline(__always)
+    public func _customContainsEquatableElement(_ element: Element) -> Bool? {
+      return _base.contains(element)
+    }
   }
 
   /// A view of a `Heap`'s elements, as a `Sequence` from the largest to
@@ -55,6 +60,11 @@ extension Heap {
     @inlinable @inline(__always)
     public var underestimatedCount: Int {
       _base.count
+    }
+
+    @inlinable @inline(__always)
+    public func _customContainsEquatableElement(_ element: Element) -> Bool? {
+      return _base.contains(element)
     }
   }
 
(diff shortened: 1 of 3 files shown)


lorentey:
I'm closing this without merging. We decided to get rid of the ordered views -- I was and remain unconvinced that they serve any useful purpose in practice that cannot be better achieved with other means. (They were over-elaborate equivalents of `sequence(heap, next: { $0.popMin() })`; but perhaps more importantly, they did not behave like views do in the stdlib. The hidden cost of having to copy and mutate a hidden copy of the original heap made them a non-starter.) If I'm proved wrong, we will have plenty of chance to re-add them in a future release.

Adding a `contains(_:)` method is not necessarily a bad idea, and we may revisit it in the future. But it alone is probably not that useful -- it should also come with some sort of equivalent to `firstIndex(of:)`/`lastIndex(of:)`, and that opens up a can of worms I don't think we're ready to tackle yet. The question is mostly about whether concrete use cases want to be able to find and remove previously added items, or update their value without having to suffer an O(n) search. Determining whether or not a particular item is already in the heap is probably not that interesting on its own, and if it was, then it likely would be better achieved by maintaining a `Set` of existing items on the side. The proposed `contains` implementation is not immediately viable, so there is little reason to keep this PR open just in case we'll need a `contains` later.

Let's keep an eye on user feedback once 1.1.0 is out, and evolve `Heap`'s APIs based on that. If feedback indicates that e.g. potential adopters are interested in a priority queue implementation that supports out-of-order retrieval, then we should design an update that directly covers their use case, rather than trying to imagine what it might be.

PR #159, opened 2022-08-16
+OrderedSet add filter #158

### LINKED ISSUE #158: Provide an OrderedSet.filter operation; currently the Sequence one is used thus an Array is returned
I'm moving some code to use `OrderedSet` and noticed it is missing a `filter` operation.

Previous code:

```swift
            // 1) we remove any registrations that it hosted
            let registrations: Set<VersionedRegistration> = self._registrations.removeValue(forKey: key) ?? []
            let remainingRegistrations: Set<VersionedRegistration> = registrations.filter { registration in
                registration.actorID.uniqueNode != node
            }
```

this was ok, since we call the filter on:

```
extension Set {
    @available(swift 4.0)
    @inlinable public func filter(_ isIncluded: (Element) throws -> Bool) rethrows -> Set<Element>
}
```

but it seems that OrderedSet is missing this specialization, so we fallback to calling the `Sequence`'s impl:

```
    @inlinable public func filter(_ isIncluded: (Self.Element) throws -> Bool) rethrows -> [Self.Element]
```

returning an array -- which is a bit meh, since we have to convert back into the `OrderedSet`.

### Feature Request

Please provide a `filter` similar to how `Set` does that keeps the collection type `OrderedSet -> OrderedSet`

ktoso:
Resolves #158

This adds a missing operation on `OrderedSet`: filter that keeps the return type an `OrderedSet`.
The dictionary in this package already had such operation, so I followed its prior art.

One could debate if we want to preallocate the size as "same" of the current set before filtering or if not etc... Open to suggestions, it's a bit tricky to know if the result of the filter will be almost empty, or almost full after all.

### Checklist
    - dictionary filter didn't seem to have one either
   - docc

### DIFF
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet.swift
@@ -514,6 +514,7 @@ extension OrderedSet {
     for element in self where try isIncluded(element) {
       result._appendNew(element)
     }
+    result._checkInvariants()
     return result
   }
 }
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet.swift
@@ -494,3 +494,26 @@ extension OrderedSet {
     return _elements.remove(at: index)
   }
 }
+
+extension OrderedSet {
+  /// Returns a new ordered set containing the values pairs of the ordered set
+  /// that satisfy the given predicate.
+  ///
+  /// - Parameter isIncluded: A closure that takes a value as its
+  ///   argument and returns a Boolean value indicating whether the value
+  ///   should be included in the returned dictionary.
+  ///
+  /// - Returns: An ordered set of the values that `isIncluded` allows.
+  ///
+  /// - Complexity: O(`count`)
+  @inlinable
+  public func filter(
+          _ isIncluded: (Element) throws -> Bool
+  ) rethrows -> Self {
+    var result: OrderedSet = Self(minimumCapacity: _minimumCapacity)
+    for element in self where try isIncluded(element) {
+      result._appendNew(element)
+    }
+    return result
+  }
+}
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet.swift
@@ -508,7 +508,7 @@ extension OrderedSet {
   /// - Complexity: O(`count`)
   @inlinable
   public func filter(
-          _ isIncluded: (Element) throws -> Bool
+    _ isIncluded: (Element) throws -> Bool
   ) rethrows -> Self {
     var result: OrderedSet = Self(minimumCapacity: _minimumCapacity)
     for element in self where try isIncluded(element) {
--- a/Tests/OrderedCollectionsTests/OrderedSet/OrderedSetTests.swift
+++ b/Tests/OrderedCollectionsTests/OrderedSet/OrderedSetTests.swift
@@ -1360,4 +1360,22 @@ class OrderedSetTests: CollectionTestCase {
       }
     }
   }
+
+
+  func test_filter() {
+    let items = (0 ..< 100)
+    let s = OrderedSet(items)
+
+    var c = 0
+    let s2 = s.filter { item in
+      c += 1
+      return item.isMultiple(of: 2)
+    }
+    expectEqual(c, 100)
+    expectEqualElements(s, items)
+
+    expectEqualElements(s2, (0 ..< 50).compactMap { key in
+      return (2 * key)
+    })
+  }
 }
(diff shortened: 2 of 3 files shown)


ktoso:
@swift-ci please smoke test

ktoso:
@swift-ci please test

Sources/OrderedCollections/OrderedSet/OrderedSet.swift
```diff
@@ -494,3 +494,26 @@ extension OrderedSet {
...
+  /// - Complexity: O(`count`)
+  @inlinable
+  public func filter(
+          _ isIncluded: (Element) throws -> Bool
+  ) rethrows -> Self {
+    var result: OrderedSet = Self(minimumCapacity: _minimumCapacity)
+    for element in self where try isIncluded(element) {
+      result._appendNew(element)
+    }
+    return result
```
lorentey, line 517: Nit: let's call `result._checkInvariants()` before returning the result.

Sources/OrderedCollections/OrderedSet/OrderedSet.swift
```diff
@@ -494,3 +494,26 @@ extension OrderedSet {
...
+  ///   argument and returns a Boolean value indicating whether the value
+  ///   should be included in the returned dictionary.
+  ///
+  /// - Returns: An ordered set of the values that `isIncluded` allows.
+  ///
+  /// - Complexity: O(`count`)
+  @inlinable
+  public func filter(
+          _ isIncluded: (Element) throws -> Bool
+  ) rethrows -> Self {
```
lorentey, line 512: This implementation can be improved by using a temporary bitset, as in [`Set.filter`](https://github.com/apple/swift/blob/main/stdlib/public/core/NativeSet.swift#L722-L736) -- I expect this will boost performance by 2x or more, by eliminating the need to repeatedly resize the hash table.

(This does not need to be implemented before this lands.)

Sources/OrderedCollections/OrderedSet/OrderedSet.swift
ktoso: Gotcha, thanks 👍

Sources/OrderedCollections/OrderedSet/OrderedSet.swift
ktoso: Hm, I poked a little bit at it but we end up having to pull in quite a bunch of new uncheckedElement, _bucketCounts etc I'm not super sure if I'm implementing them right -- so leaving this as future work I guess. Happy to make a ticket for it after this lands.

Tests/OrderedCollectionsTests/OrderedSet/OrderedSetTests.swift
```diff
@@ -1360,4 +1360,22 @@ class OrderedSetTests: CollectionTestCase {
       }
     }
   }
+
+
+  func test_filter() {
+    let items = (0 ..< 100)
```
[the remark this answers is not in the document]
ktoso: Nice, thanks for proposing the property testing approach, missed that you have those here, nice 👍 

Also added the type test.

ktoso:
@swift-ci please test

Sources/OrderedCollections/OrderedSet/OrderedSet.swift
```diff
@@ -494,3 +494,27 @@ extension OrderedSet {
...
+  /// - Parameter isIncluded: A closure that takes a value as its
+  ///   argument and returns a Boolean value indicating whether the value
+  ///   should be included in the returned dictionary.
+  ///
+  /// - Returns: An ordered set of the values that `isIncluded` allows.
+  ///
+  /// - Complexity: O(`count`)
+  @inlinable
+  public func filter(
+          _ isIncluded: (Element) throws -> Bool
```
xwu, line 511:
```suggestion
    _ isIncluded: (Element) throws -> Bool
```

ktoso:
Fixed the formatting, thanks @xwu -- though on that note I'd really suggest adding a CI step to verify formatting.

Manually maintaining and checking these isn't tangible. We do this in all server projects we maintain with a "soundness" CI step: https://github.com/apple/swift-distributed-actors/blob/main/scripts/soundness.sh Feel free to take those scripts (`validate_format`), we're using them in all server side repos we maintain.

ktoso:
@swift-ci please test

Sources/OrderedCollections/OrderedSet/OrderedSet.swift
lorentey: Oh, the stdlib code wouldn't directly translate -- that's just illustrating the concept. We have a separate bitset implementation within this code base, so we'd need to adapt the algorithm to use that instead. ([OrderedSet.symmetricDifference](https://github.com/apple/swift-collections/blob/2ce28a9eaac3a03c7403b09aae10fea3273de193/Sources/OrderedCollections/OrderedSet/OrderedSet%2BPartial%20SetAlgebra%2BOperations.swift#L290-L314) is an example.)

I'm happy to do this as a followup.

ktoso: Followup: https://github.com/apple/swift-collections/pull/163

PR #163, opened 2022-08-18
=OrderedSet.filter Attempt to optimize filter impl

ktoso:
Follow up to https://github.com/apple/swift-collections/pull/159

This is pretty weird, the "optimized" version is slower, and some of the results are pretty suspicious.
Might want to have a look what could be off here @lorentey

I included all the results in the PR so you can browse them; this was run on an M1 macbook pro.

Screenshot 2022-08-18 at 10 30 12 AM
Screenshot 2022-08-18 at 10 30 20 AM


lorentey:
Getting rid of the hash lookups during the construction of the bitmap gets us to a comfortable place:
01 OrderedSet Int  filter keeping 100%
02 OrderedSet Int  filter keeping 75%
03 OrderedSet Int  filter keeping 50%
04 OrderedSet Int  filter keeping 25%
05 OrderedSet Int  filter keeping 0%

lorentey:
@swift-ci test

Benchmarks/Benchmarks/OrderedSetBenchmarks.swift
```diff
@@ -269,6 +269,25 @@ extension Benchmark {
...
+          timer.measure {
+            r = set.filter { $0 % 100 < percent }
+          }
+          let div = input.count / 100
+          let rem = input.count % 100
+          precondition(r.count == percent * div + Swift.min(rem, percent))
+          blackHole(r)
+        }
+      }
+    }
```
ktoso, line 289: I like your benchmark suite a lot btw, very nice work 👍

ktoso (commented):
Very nice, thank you for the follow up here @lorentey !

Ready to merge then AFAICS 👍

PR #165, opened 2022-08-31
Force-inline _modify accessors to work around a performance issue

### LINKED ISSUE #164: Deque subscript _modify accessor heap allocates
Consider the following simple function (the operation is not useful, it's just intended to be complex enough to defeat being optimized away and simple enough to make the generated assembly simple):

```swift
func twiddle(index: Int, twiddle: UInt8, deque: inout Deque<UInt8>) {
    deque[index] ^= twiddle
}
```

When compiled in release mode using swift-collections 1.0.2 and current `main` ([`c3fdcf7`](https://github.com/apple/swift-collections/commit/c3fdcf746e6cd44bf9c74ad93ac80f3677e91234)), this generates the following assembly on my arm64 Mac:

```
_$s18alloc_modify_repro10DequeReproV7twiddle5indexAD5dequeySi_s5UInt8V0D6Module0D0VyAHGztFZTf4nnnd_n:        // function signature specialization <Arg[3] = Dead> of static alloc_modify_repro.DequeRepro.twiddle(index: Swift.Int, twiddle: Swift.UInt8, deque: inout DequeModule.Deque<Swift.UInt8>) -> ()
sub        sp, sp, #0x40                               ; CODE XREF=_$s18alloc_modify_repro10DequeReproV4mainyyFZTf4d_n+208
stp        x20, x19, [sp, #0x20]
stp        fp, lr, [sp, #0x30]
add        fp, sp, #0x30
mov        x19, x1
mov        x1, x0                                      ; argument #2 for method _$s11DequeModule0A0VyxSiciMs5UInt8V_Tg5
mov        x0, sp                                      ; argument #1 for method _$s11DequeModule0A0VyxSiciMs5UInt8V_Tg5
mov        x20, x2
bl         _$s11DequeModule0A0VyxSiciMs5UInt8V_Tg5     ; generic specialization <Swift.UInt8> of DequeModule.Deque.subscript.modify : (Swift.Int) -> A
mov        x8, x0
ldrb       w9, [x1]
eor        w9, w9, w19
strb       w9, [x1]
mov        x0, sp
mov        w1, #0x0
blr        x8
ldp        fp, lr, [sp, #0x30]
ldp        x20, x19, [sp, #0x20]
add        sp, sp, #0x40
ret
```

As you can see, `_modify` has not been inlined but it has been specialized. As a result, the call to the continuation has been split and delayed (visible as `blr x8`).

`_modify` has compiled to:

```
_$s11DequeModule0A0VyxSiciMs5UInt8V_Tg5:        // generic specialization <Swift.UInt8> of DequeModule.Deque.subscript.modify : (Swift.Int) -> A
stp        x24, x23, [sp, #-0x40]!                     ; CODE XREF=_$s18alloc_modify_repro10DequeReproV7twiddle5indexAD5dequeySi_s5UInt8V0D6Module0D0VyAHGztFZTf4nnnd_n+32
stp        x22, x21, [sp, #0x10]
stp        x20, x19, [sp, #0x20]
stp        fp, lr, [sp, #0x30]
add        fp, sp, #0x30
mov        x21, x1
mov        x22, x0
mov        w0, #0x28                                   ; argument "size" for method imp___stubs__malloc
bl         imp___stubs__malloc                         ; malloc
str        x0, [x22]
stp        x21, x20, [x0]
tbnz       x21, 0x3f, loc_100012600

mov        x19, x0
ldr        x23, [x20]
mov        x0, #0x0
bl         _$sSo10HeapObjectVMa                        ; type metadata accessor for __C.HeapObject
ldr        x8, [x23, #0x18]
cmp        x8, x21
b.le       loc_100012604

add        x22, x19, #0x20
mov        x0, x23                                     ; argument "arg0" for method imp___stubs__swift_isUniquelyReferenced_nonNull_native
bl         imp___stubs__swift_isUniquelyReferenced_nonNull_native ; swift_isUniquelyReferenced_nonNull_native
tbnz       w0, 0x0, loc_1000125b0

bl         _$s11DequeModule0A0V8_StorageV15_makeUniqueCopyyyFs5UInt8V_Tg5 ; generic specialization <Swift.UInt8> of DequeModule.Deque._Storage._makeUniqueCopy() -> ()

loc_1000125b0:
ldr        x8, [x20]                                   ; CODE XREF=_$s11DequeModule0A0VyxSiciMs5UInt8V_Tg5+88
ldr        x9, [x8, #0x20]!
str        x9, [x19, #0x10]
add        x9, x9, x21
ldur       x10, [x8, #-0x10]
str        x10, [x19, #0x18]
cmp        x9, x10
csel       x10, xzr, x10, lt
sub        x9, x9, x10
add        x8, x8, x9
ldrb       w8, [x8, #0x8]
strb       w8, [x19, #0x20]
adr        x0, #0x100012608
nop
mov        x1, x22
ldp        fp, lr, [sp, #0x30]
ldp        x20, x19, [sp, #0x20]
ldp        x22, x21, [sp, #0x10]
ldp        x24, x23, [sp], #0x40
ret

loc_100012600:
brk        #0x1                                        ; CODE XREF=_$s11DequeModule0A0VyxSiciMs5UInt8V_Tg5+44

loc_100012604:
brk        #0x1                                        ; CODE XREF=_$s11DequeModule0A0VyxSiciMs5UInt8V_Tg5+72
```

Notice the call to `malloc` at the start of the function. This call drastically dominates the performance of `_modify`, and it makes `Deque` extremely difficult to use in performance sensitive code.

### Information

- **Package version:** 1.0.2 or `main` ([`c3fdcf7`](https://github.com/apple/swift-collections/commit/c3fdcf746e6cd44bf9c74ad93ac80f3677e91234))
- **Platform version:** 21G83
- **Swift version:** `swift-driver version: 1.45.2 Apple Swift version 5.6.1 (swiftlang-5.6.0.323.66 clang-1316.0.20.12) Target: arm64-apple-macosx12.0`

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] I've searched for [existing GitHub issues](https://github.com/apple/swift-collections/issues).

### Steps to Reproduce
Use the above code sample in an empty Swift project that depends on swift-collections.

### Expected behavior
No allocations in `_modify`.

### Actual behavior
A direct call to `malloc` and a subsequent call to `free`.

lorentey:
As @Lukasa discovered in #164, the `_modify` accessor in `Deque`'s index subscript tends to call malloc unless it happens to get inlined into the caller. Try to prevent this performance bottleneck by force-inlining this accessor.

To mitigate potential code size issues, move the preparation/finalization steps into separate functions that aren't force-inlined.

Resolves #164.

### Checklist

### DIFF
--- a/Benchmarks/Benchmarks/DequeBenchmarks.swift
+++ b/Benchmarks/Benchmarks/DequeBenchmarks.swift
@@ -172,6 +172,36 @@ extension Benchmark {
     }
 
     self.add(
+      title: "Deque<Int> modify through subscript (contiguous)",
+      input: ([Int], [Int]).self
+    ) { input, lookups in
+      return { timer in
+        var deque = Deque(input)
+        timer.measure {
+          for i in lookups {
+            deque[i] += 1
+          }
+        }
+        blackHole(deque)
+      }
+    }
+
+    self.add(
+      title: "Deque<Int> modify through subscript (discontiguous)",
+      input: ([Int], [Int]).self
+    ) { input, lookups in
+      return { timer in
+        var deque = Deque(discontiguous: input)
+        timer.measure {
+          for i in lookups {
+            deque[i] += 1
+          }
+        }
+        blackHole(deque)
+      }
+    }
+
+    self.add(
       title: "Deque<Int> random swaps (contiguous)",
       input: [Int].self
     ) { input in
--- a/Sources/DequeModule/Deque+Collection.swift
+++ b/Sources/DequeModule/Deque+Collection.swift
@@ -359,22 +359,33 @@ extension Deque: RandomAccessCollection {
         handle.ptr(at: slot).pointee = newValue
       }
     }
+    @inline(__always) // https://github.com/apple/swift-collections/issues/164
     _modify {
       precondition(index >= 0 && index < count, "Index out of bounds")
-      _storage.ensureUnique()
-      // We technically aren't supposed to escape storage pointers out of a
-      // managed buffer, so we escape a `(slot, value)` pair instead, leaving
-      // the corresponding slot temporarily uninitialized.
-      var (slot, value) = _storage.update { handle -> (_Slot, Element) in
-        let slot = handle.slot(forOffset: index)
-        return (slot, handle.ptr(at: slot).move())
-      }
+      var (slot, value) = _prepareForModify(at: index)
       defer {
-        _storage.update { handle in
-          handle.ptr(at: slot).initialize(to: value)
-        }
+        _finalizeModify(slot, value)
       }
       yield &value
+    }
+  }
+
+  @inlinable
+  internal mutating func _prepareForModify(at index: Int) -> (_Slot, Element) {
+    _storage.ensureUnique()
+    // We technically aren't supposed to escape storage pointers out of a
+    // managed buffer, so we escape a `(slot, value)` pair instead, leaving
+    // the corresponding slot temporarily uninitialized.
+    return _storage.update { handle in
+      let slot = handle.slot(forOffset: index)
+      return (slot, handle.ptr(at: slot).move())
+    }
+  }
+
+  @inlinable
+  internal mutating func _finalizeModify(_ slot: _Slot, _ value: Element) {
+    _storage.update { handle in
+      handle.ptr(at: slot).initialize(to: value)
     }
   }
 
--- a/Sources/OrderedCollections/HashTable/_HashTable.swift
+++ b/Sources/OrderedCollections/HashTable/_HashTable.swift
@@ -170,6 +170,7 @@ extension _HashTable {
   @inlinable
   internal var header: Header {
     get { _storage.header }
+    @inline(__always) // https://github.com/apple/swift-collections/issues/164
     nonmutating _modify { yield &_storage.header }
   }
 
--- a/Sources/OrderedCollections/OrderedDictionary/OrderedDictionary+Elements.swift
+++ b/Sources/OrderedCollections/OrderedDictionary/OrderedDictionary+Elements.swift
@@ -35,6 +35,7 @@ extension OrderedDictionary {
     get {
       Elements(_base: self)
     }
+    @inline(__always) // https://github.com/apple/swift-collections/issues/164
     _modify {
       var elements = Elements(_base: self)
       self = Self()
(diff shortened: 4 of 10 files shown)


lorentey:
@swift-ci test

lorentey:
It turns out we did not have any benchmarks exercising this accessor. The improvement is pretty sizable:

01 Deque Int  modify through subscript (discontiguous)
02 Deque Int  modify through subscript (contiguous)

`malloc` is quite fast though! The extra 10ns does dominate this benchmark, but in higher-level code it can often be swamped by the actual work.

lorentey:
@swift-ci test

lorentey:
@swift-ci test

lorentey:
@swift-ci test

Lukasa:
@lorentey Is that a 10ns slowdown? The y axis is logarithmic, so isn't this a 10x improvement? (Eyeballing logarithmic scales is hard but we appear to be about 1/3 of the way through the 10ns to 100ns branch in the "before" which would be about 20ns, as compared to also about 1/3 of the way through the 1ns to 10ns branch in the "after", which would be about 2ns.)

Regardless, I agree the gain is great. I also think this is a best-case scenario for malloc, as the allocator is uncontended. I'd be curious if we can run the benchmark concurrently in multiple threads. That should slow it down due to increased synchronization overhead between threads and produce an even more marked gain.

(I realise now that the benchmark tool is OSS, so feel free to tell me to buzz off and get my own data :wink:)

lorentey:
Ah good point, I haven't considered allocations getting swamped. Is there typically no thread-local fast path?

The benchmarking facility is very intentionally single-threaded (benchmarking concurrent code well is tricky), but it wouldn't be too hard to introduce some artificial background task as an experiment. I'm not interested enough to do it myself right now though. :-P

lorentey:
Yes you are right, the absolute values are closer to 2ns vs 20ns, so the measured cost for malloc/free is 18ns, not 10ns.

Lukasa:
> Ah good point, I haven't considered allocations getting swamped. Is there typically no thread-local fast path?

To bring us home, the answer is "sometimes, it depends on the allocator". I can't speak to the Darwin allocator which I don't know well enough, but glibc _does_ have a thread-local fast path at the cost of massively increasing memory fragmentation, while allocators like jemalloc or tcmalloc will have a thread-local fast path that occasionally attempts to steal memory from across threads, causing coordination hits. In general, adding tcmalloc to a multithreaded program on Linux will greatly reduce its memory usage at the cost of mildly increasing its CPU usage.

Catfish-Man (commented):
Changes look reasonable to me :)

Catfish-Man:
Darwin has a lockless fast path, but it's disabled in daemons for fragmentation reasons

PR #173, opened 2022-09-07
Consistency in file names

rowungiles:
A small number of files have a naming structure which appears to be inconsistent with the majority of files in the project. This PR aligns the filenames.

### Checklist


lorentey:
`+` isn't meant to stand in for a space character -- I use it as a shorthand for files that extend a type with functionality.

Do you have a particular issue with the original file names?

rowungiles:
That loose Obj-C naming convention makes sense, it's my preferred way to represent extension of functionality too. I didn't intend `+` to be an arbitrary replacement for space.

I think some of the file names could be more consistent, for example
`OrderedSet.UnorderedView Tests.swift` which relates to `OrderedSet+UnorderedView.swift` in the test name it uses a `.` but the implementation uses `+`. Or `OrderedSet Diffing Tests.swift` and `OrderedSetTests.swift` in the former, `Tests` is attached to the name, but in the latter it isn't.

The naming conventions I've always seen are file names match the type name unless functionality is being extended then using the `+` pattern. I'm not sure I've seen a convention for conveying additional information in the file name. I had a quick look through other `/apple` open source and it appears `_` may be being used instead of a space? 

So in the example `OrderedSet+Partial MutableCollection.swift` I proposed conveying extension of functionality twice but would `OrderedSet+Partial_MutableCollection.swift` better convey intent?

lorentey:
Oh yeah, I keep wanting to name files defining nested types with dots (as in `OrderedDictionary.Elements.swift`), but then I invariably get annoyed at the way Xcode insists on sorting them (e.g., before `OrderedDictionary.swift`), so I end up using the plus sign as a compromise. The dots remained in the test suite in the case of `OrderedSet.UnorderedView Tests.swift`, where the ASCII sort order happened to work well. This is fine; renaming it isn't worth the trouble.

Re spaces: it's 2022. Spaces in filenames are neither scary nor troublesome these days. I feel quite strongly that there is no valid reason to avoid using them, and I'm _quite_ happy to use this package to root out any broken tools. The spaces are here to stay. 😇

rowungiles:
The spaces do break some tools, but that's not a problem for this package to solve. 

Ok, if you don't think there is an improvement to be made here, I'll close the PR.

PR #182, opened 2022-09-21
Persistent collections updates (part 7)

lorentey:
This fixes the `Iterator` type (slowing it down by 35% or so, unfortunately), and starts work on `PersistentSet`. (Including (untested) structural isSubset and isDisjoint tests, but no structural merges yet.)

### Checklist

### DIFF
--- a/Sources/PersistentCollections/Node/_Node+isDisjoint.swift
+++ b/Sources/PersistentCollections/Node/_Node+isDisjoint.swift
@@ -76,6 +76,20 @@ extension _Node {
   ) -> Bool {
     // Beware, self might be on a compressed path
     assert(isCollisionNode)
+    if other.isCollisionNode {
+      return read { l in
+        other.read { r in
+          let lh = l.collisionHash
+          let rh = r.collisionHash
+          guard lh == rh else { return true }
+          let litems = l.reverseItems
+          let ritems = r.reverseItems
+          return litems.allSatisfy { li in
+            !ritems.contains { ri in li.key == ri.key }
+          }
+        }
+      }
+    }
     return read {
       let items = $0.reverseItems
       let hash = $0.collisionHash
--- a/Sources/PersistentCollections/Node/_Node+isDisjoint.swift
+++ b/Sources/PersistentCollections/Node/_Node+isDisjoint.swift
@@ -35,7 +35,7 @@ extension _Node {
         for bucket in l.itemMap.intersection(r.itemMap) {
           let lslot = l.itemMap.slot(of: bucket)
           let rslot = r.itemMap.slot(of: bucket)
-          guard l[item: lslot].key == r[item: rslot].key else { return false }
+          guard l[item: lslot].key != r[item: rslot].key else { return false }
         }
         for bucket in l.itemMap.intersection(r.childMap) {
           let lslot = l.itemMap.slot(of: bucket)
--- a/Sources/PersistentCollections/Node/_Node+isDisjoint.swift
+++ b/Sources/PersistentCollections/Node/_Node+isDisjoint.swift
@@ -1,0 +1,87 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Collections open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+extension _Node {
+  /// Returns true if `self` contains a disjoint set of keys than `other`.
+  /// Otherwise, returns false.
+  @inlinable @inline(never)
+  internal func isDisjoint<Value2>(
+    _ level: _Level,
+    with other: _Node<Key, Value2>
+  ) -> Bool {
+    if self.raw.storage === other.raw.storage { return count == 0 }
+
+    if self.isCollisionNode {
+      return _isDisjointCollision(level, with: other)
+    }
+    if other.isCollisionNode {
+      return other._isDisjointCollision(level, with: other)
+    }
+
+    return self.read { l in
+      other.read { r in
+        let lmap = l.itemMap.union(l.childMap)
+        let rmap = r.itemMap.union(r.childMap)
+        if lmap.isDisjoint(with: rmap) { return true }
+
+        for bucket in l.itemMap.intersection(r.itemMap) {
+          let lslot = l.itemMap.slot(of: bucket)
+          let rslot = r.itemMap.slot(of: bucket)
+          guard l[item: lslot].key == r[item: rslot].key else { return false }
+        }
+        for bucket in l.itemMap.intersection(r.childMap) {
+          let lslot = l.itemMap.slot(of: bucket)
+          let hash = _Hash(l[item: lslot].key)
+          let rslot = r.childMap.slot(of: bucket)
+          let found = r[child: rslot].containsKey(
+            level.descend(),
+            l[item: lslot].key,
+            hash)
+          if found { return false }
+        }
+        for bucket in l.childMap.intersection(r.itemMap) {
+          let lslot = l.childMap.slot(of: bucket)
+          let rslot = r.itemMap.slot(of: bucket)
+          let hash = _Hash(r[item: rslot].key)
+          let found = l[child: lslot].containsKey(
+            level.descend(),
+            r[item: rslot].key,
+            hash)
+          if found { return false }
+        }
+        for bucket in l.childMap.intersection(r.childMap) {
+          let lslot = l.childMap.slot(of: bucket)
+          let rslot = r.childMap.slot(of: bucket)
+          guard
+            l[child: lslot].isDisjoint(level.descend(), with: r[child: rslot])
+          else { return false }
+        }
+        return true
+      }
+    }
+  }
+
+  @inlinable @inline(never)
+  internal func _isDisjointCollision<Value2>(
+    _ level: _Level,
+    with other: _Node<Key, Value2>
+  ) -> Bool {
+    // Beware, self might be on a compressed path
+    assert(isCollisionNode)
+    return read {
+      let items = $0.reverseItems
+      let hash = $0.collisionHash
+      return items.indices.allSatisfy {
+        !other.containsKey(level, items[$0].key, hash)
+      }
+    }
+  }
+}
--- a/Sources/PersistentCollections/Node/_Node+isEqual.swift
+++ b/Sources/PersistentCollections/Node/_Node+isEqual.swift
@@ -10,40 +10,46 @@
 //===----------------------------------------------------------------------===//
 
 // TODO: `Equatable` needs more test coverage, apart from hash-collision smoke test
-extension _Node: Equatable where Value: Equatable {
+extension _Node {
   @inlinable
-  static func == (left: _Node, right: _Node) -> Bool {
-    if left.raw.storage === right.raw.storage { return true }
+  internal func isEqual(
+    to other: _Node,
+    by areEquivalent: (Element, Element) -> Bool
+  ) -> Bool {
+    if self.raw.storage === other.raw.storage { return true }
 
-    guard left.count == right.count else { return false }
+    guard self.count == other.count else { return false }
 
-    if left.isCollisionNode {
-      guard right.isCollisionNode else { return false }
-      return left.read { lhs in
-        right.read { rhs in
+    if self.isCollisionNode {
+      guard other.isCollisionNode else { return false }
+      return self.read { lhs in
+        other.read { rhs in
           let l = lhs.reverseItems
           let r = rhs.reverseItems
           guard l.count == r.count else { return false }
           for i in l.indices {
-            guard r.contains(where: { $0 == l[i] }) else { return false }
+            guard r.contains(where: { areEquivalent($0, l[i]) })
+            else { return false }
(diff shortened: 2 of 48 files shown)


msteindorfer:
In general, in case keys are objects, should be consider checking for object identity (i.e., `===`) in the persistent collections before calling `==`?

Can this be achieved in a single method implementation, or would one need to duplicate the method and add a constraint like `Key: AnyObject`?

Sources/PersistentCollections/Node/_Node+isSubset.swift
```diff
@@ -0,0 +1,71 @@
...
+      // Beware, self might be on a compressed path
+      return read {
+        let items = $0.reverseItems
+        let hash = $0.collisionHash
+        return items.indices.allSatisfy {
+          // FIXME: This will repeatedly & unnecessarily call `collisionHash`
+          other.containsKey(level, items[$0].key, hash)
+        }
+      }
+    }
```
msteindorfer, line 33: What was the rationale to extract `_isDisjointCollision` but keep the code inlined for `isSubset`?

Sources/PersistentCollections/Node/_Node+isSubset.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    guard !other.isCollisionNode else { return false }
+
+    return self.read { l in
+      other.read { r in
+        guard l.childMap.isSubset(of: r.childMap) else { return false }
+        guard l.itemMap.isSubset(of: r.itemMap.union(r.childMap)) else {
+          return false
+        }
```
msteindorfer, line 42: 👍

Sources/PersistentCollections/Node/_Node+isSubset.swift
```diff
@@ -0,0 +1,71 @@
...
+//===----------------------------------------------------------------------===//
+
+extension _Node {
+  /// Returns true if `self` contains a subset of the keys in `other`.
+  /// Otherwise, returns false.
+  @inlinable @inline(never)
+  internal func isSubset<Value2>(
+    _ level: _Level,
+    of other: _Node<Key, Value2>
+  ) -> Bool {
```
msteindorfer, line 19: Consider adding benchmarks for `isSubset` and other structural operations.

Structural operations with persistent data structures have an inherent benefit over performing those operations on a logical level. Hence, it would be great to highlight those differences with benchmarks.

Sources/PersistentCollections/Node/_Stack.swift
```diff
@@ -0,0 +1,115 @@
...
+/// `PersistentDictionary`.
+@usableFromInline
+@frozen
+internal struct _Stack<Element> {
+#if arch(x86_64) || arch(arm64)
+  @inlinable
+  @inline(__always)
+  internal static var capacity: Int { 13 }
+
+  @usableFromInline
```
msteindorfer, line 22: For consistency with `#else`?

```suggestion
  // xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xxxx
  @usableFromInline
```

Sources/PersistentCollections/Node/_Stack.swift
```diff
@@ -0,0 +1,115 @@
...
+  @inline(__always)
+  internal static var capacity: Int { 7 }
+
+  // xxxxx xxxxx xxxxx xxxxx xxxxx xxxxx xx
+  @usableFromInline
+  internal var _contents: (
+    Element, Element, Element, Element,
+    Element, Element, Element
+  )
+#endif
```
msteindorfer, line 40: Is `hashValue` in Swift on 32-bit systems 32 bits as well?

Sources/PersistentCollections/PersistentDictionary/PersistentDictionary+Filter.swift
```diff
@@ -17,14 +17,10 @@ extension PersistentDictionary {
     var result: PersistentDictionary = [:]
     for item in self {
       guard try isIncluded(item) else { continue }
-      // FIXME: We could recover the key's hash from self.
-      // (As many bits of it as we need to insert it into the new tree.)
-      // However, this requires a series of `UInt32._bit(ranked:)` invocations
-      // that may or may not be meaningfully better than simply rehashing the
-      // key.
+      // FIXME: We could do this as a structural transformation.
```
msteindorfer, line 20: 👍

msteindorfer: Consider that structural formations must deal with compaction in order to guarantee a canonical representation, which slightly complicates the matter.

Sources/PersistentCollections/PersistentDictionary/PersistentDictionary+MapValues.swift
```diff
@@ -22,17 +22,13 @@ extension PersistentDictionary {
   public func compactMapValues<T>(
     _ transform: (Value) throws -> T?
   ) rethrows -> PersistentDictionary<Key, T> {
+    // FIXME: We could do this as a structural transformation.
```
msteindorfer, line 25: 👍

msteindorfer: Consider that structural formations must deal with compaction in order to guarantee a canonical representation, which slightly complicates the matter.

Sources/PersistentCollections/PersistentDictionary/PersistentDictionary+Merge.swift
```diff
@@ -15,6 +15,7 @@ extension PersistentDictionary {
     _ keysAndValues: __owned S,
     uniquingKeysWith combine: (Value, Value) throws -> Value
   ) rethrows where S.Element == (Key, Value) {
+    // FIXME: Do a structural merge
```
msteindorfer, line 18: 👍

lorentey:
There isn't really a good way to change behavior based on properties of a generic type argument (other than using the type constraints required in the declaration) 

Doing things like `foo is AnyObject` can backfire pretty badly if the check survives into the generated code (e.g. when foo is of a type that bridges with Objective C).

The === shortcut is best implemented within the Equatable.== implementation of the Key type, like we do for the collection types themselves.

Sources/PersistentCollections/Node/_Node+isDisjoint.swift
```diff
@@ -0,0 +1,87 @@
...
+    return self.read { l in
+      other.read { r in
+        let lmap = l.itemMap.union(l.childMap)
+        let rmap = r.itemMap.union(r.childMap)
+        if lmap.isDisjoint(with: rmap) { return true }
+
+        for bucket in l.itemMap.intersection(r.itemMap) {
+          let lslot = l.itemMap.slot(of: bucket)
+          let rslot = r.itemMap.slot(of: bucket)
+          guard l[item: lslot].key == r[item: rslot].key else { return false }
```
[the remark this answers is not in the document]
lorentey: Yep, good eye!

Sources/PersistentCollections/Node/_Node+isEqual.swift
```diff
@@ -10,40 +10,46 @@
...
 // TODO: `Equatable` needs more test coverage, apart from hash-collision smoke test
-extension _Node: Equatable where Value: Equatable {
+extension _Node {
   @inlinable
-  static func == (left: _Node, right: _Node) -> Bool {
-    if left.raw.storage === right.raw.storage { return true }
+  internal func isEqual(
+    to other: _Node,
+    by areEquivalent: (Element, Element) -> Bool
+  ) -> Bool {
```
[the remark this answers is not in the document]
lorentey: The cost is usually pretty minimal with these -- this is a non-escaping closure, so it can and will be stack-allocated when possible! Inlining of the closure may trigger as well, but I don't really have an intuition whether it is likely to happen here (tbh I don't even know if the compiler would be willing to specialize functions for specific closures like that).

If I had some free time, I would experiment with switching `_Node.Element` from a tuple to an internal frozen struct, so it would be able to carry an (easily inlinable) Equatable conformance.

msteindorfer (approved):
LGTM

Sources/PersistentCollections/Node/_Node+isDisjoint.swift
```diff
@@ -0,0 +1,87 @@
...
+//===----------------------------------------------------------------------===//
+
+extension _Node {
+  /// Returns true if `self` contains a disjoint set of keys than `other`.
+  /// Otherwise, returns false.
+  @inlinable @inline(never)
+  internal func isDisjoint<Value2>(
+    _ level: _Level,
+    with other: _Node<Key, Value2>
+  ) -> Bool {
```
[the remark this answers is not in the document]
lorentey: Yep. 

Note: Benchmarking these well is a challenge -- the state space is so large, and the behavior is so dependent on the random hash seed, that it's hard to come up with something that truly demonstrates practical benefits.

Sources/PersistentCollections/Node/_Node+isDisjoint.swift
```diff
@@ -0,0 +1,87 @@
...
+    if self.raw.storage === other.raw.storage { return count == 0 }
+
+    if self.isCollisionNode {
+      return _isDisjointCollision(level, with: other)
+    }
+    if other.isCollisionNode {
+      return other._isDisjointCollision(level, with: other)
+    }
```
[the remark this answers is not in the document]
lorentey: Good idea! I'm adding this.

Sources/PersistentCollections/Node/_Node+isEqual.swift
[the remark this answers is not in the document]
lorentey: On second look, it looks I ended up changing `isSubset(of:)` and `isDisjoint(with:)` to get rid of the closure -- it superficially allows degrees of freedom that aren't actually supported. (I.e., we cannot really customize key comparisons, as that would conflict with our hashing scheme.) I'll replace it with a generic argument for Value instead.

Sources/PersistentCollections/PersistentDictionary/PersistentDictionary+MapValues.swift
lorentey: It certainly does. Trying to do this in place (for `Set.formIntersection` and the like) would probably be a fool's errand, but I think an algorithm that constructs a new instance should be reasonably doable! (This is next on my list for part 9, actually. 🤓)

Sources/PersistentCollections/Node/_Stack.swift
lorentey: Yep! `Hasher` returns exactly `Int`-sized globs of bits.

lorentey: It's a shame we have to hardwire these lengths here, by the way, and also that we have to manually list architectures instead of just [directly asking the compiler about the size of `Int`](https://forums.swift.org/t/pitch-pointer-bit-width-compile-time-conditional/59572)...

Sources/PersistentCollections/Node/_Stack.swift
lorentey: Oh oops I didn't mean to commit that doodle actually -- I just wanted to prove to myself the values are right. But it does seem like a good visual, so let's do it

lorentey:
@swift-ci test

Sources/PersistentCollections/Node/_Node+isEqual.swift
```diff
@@ -12,9 +12,9 @@
 // TODO: `Equatable` needs more test coverage, apart from hash-collision smoke test
 extension _Node {
   @inlinable
-  internal func isEqual(
-    to other: _Node,
-    by areEquivalent: (Element, Element) -> Bool
+  internal func isEqual<Value2>(
+    to other: _Node<Key, Value2>,
+    by areEquivalent: (Value, Value2) -> Bool
   ) -> Bool {
```
msteindorfer, line 18: Why is it necessary to introduce the additional generic `Value2` here?

lorentey: This isn't exercised, and there is a chance it won't be, in which case we can remove it later. (But it doesn't really have a performance drawback, so it's no big deal if it stays in.)

The reason I wanted to leave this door open is in case we want to tackle a weird asymmetry in SetAlgebra. It is customary for SetAlgebra types to provide flexible overloads for `isSubset`/`isDisjoint`/`intersection`/etc that can take more types than just `Self` -- elsewhere I usually have an unconstrained generic overload that can take any `Sequence` plus a handful of concrete overloads for select types for which I can do a more efficient implementation. (So e.g. `OrderedSet` provides not only `isSubset(of other: Self)`, but also `isSubset(of other: Set<Element>)` and `isSubset(of other: some Sequence<Element>)`.) The asymmetry I find weird is that this applies to all such operations except `Equatable.==` -- why can't I easily check if a `Set` and an `OrderedSet` contain exactly the same members?

In the case of `PersistentSet`, we currently have `isSubset(of:)` defined for `Self` and `PersistentDictionary<Element, T>.Keys` arguments. (Later on we will also need one that can take any `SetAlgebra` and one that can take any `Sequence`.) If we do end up adding an `isEqual(to:)` family of methods that generalize `==`, then we will need to this particular flavor of generic definition in `_Node` to be able to compare a persistent set to the `Keys` view of any persistent dictionary that has a matching `Key` type.

lorentey:
@swift-ci test

PR #215, opened 2022-10-19
Update CMakeLists.txt

compnerd:
Remove missing file.  Thanks to @stevenbrix for identifying this!

### DIFF
--- a/Sources/PersistentCollections/CMakeLists.txt
+++ b/Sources/PersistentCollections/CMakeLists.txt
@@ -16,7 +16,6 @@ add_library(PersistentCollections
   "Node/_HashTreeStatistics.swift"
   "Node/_Level.swift"
   "Node/_Node+Builder.swift"
-  "Node/_Node+CustomStringConvertible.swift"
   "Node/_Node+Debugging.swift"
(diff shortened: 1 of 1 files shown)


lorentey (approved):

lorentey:
Hm, this needs to go on release/1.1 too. I'll cherry-pick it.

lorentey:
@swift-ci test

lorentey:
@stevenbrix Incidentally, how did you discover this? (I hope y'all aren't using unreleased commits from `main` in production -- this repository really isn't really suitable for such use yet, as PR testing isn't nearly strong enough to catch obvious mistakes, and manual test runs only occur once in a blue moon.)

stevenbrix:
@lorenty no not using it in production, I was playing around with building the swift on windows tool chain

lorentey:
Ah, so it sounds like toolchain builds might be picking up the `main` branch of this package, rather than a stable release tag. That is not great -- the toolchain would probably be better of using a specific stable version, even in development builds. 

This isn't a huge problem, but it might unnecessary increase our work load with these sorts of trivial-but-not-caught-by-PR-testing issues. I'll make a note to look into it.

PR #265, opened 2023-03-02
build: add ARM64 to the architecture mappings for Windows

compnerd:
Windows uses `ARM64` for `CMAKE_SYSTEM_PROCESSOR` which does not match the current spellings.


lorentey (approved):

lorentey:
@swift-ci test

PR #287, opened 2023-05-19
[BitSet] Fix a thinko in BitSet.isEqualSet

lorentey:
If the interval we're checking against happens to end on a storage word boundary, then we mistakenly returned false even if the comparison should've succeeded.

This was caught by exercising the 32-bit code paths on watchOS -- one of the tests happened to use 130..<160 as the range, and 160 == 5 * 32. Add new test cases that would've caught this on 64 bit systems, too.

### Checklist


lorentey:
@swift-ci test

PR #291, opened 2023-06-03
`Rope.find` returns a bogus remainder for the end position

lorentey:
When looking for the rope position precisely at the end of the rope, `Rope.find(at:in:preferEnd:)` fails to return the right result: it correctly returns the `endIndex`, but it indicates that the position we’re looking for is at a large offset into that (nonexistent) element, instead of correctly reporting the remainder as zero:

```swift
let chunks = (0 ..< 5).map { Chunk(length: 10, value: $0) }
let rope = Rope<Chunk>(chunks)
// `rope` contains 5 chunks, with 50 components according to the metric `Chunk.Metric`.

let (index, remaining) = rope.find(at: 50, in: Chunk.Metric(), preferEnd: false)
print(index == rope.endIndex) // true, OK
print(remaining) // prints 50, should be 0
```

Predictably, this can have dire consequences in client code.

rdar://110191289

### Checklist


lorentey:
@swift-ci test

Azoy:
👍

PR #292, opened 2023-06-04
Update OrderedDictionary/OrderedSet documentation

hassila:
Update documentation to reflect that all bits used in the hashing needs to be part of the equality check for correctness (not strictly necessarily vice-versa as the current documentation reads). 

See also:

https://github.com/apple/swift/issues/66282#issuecomment-1572642416

### Checklist

### DIFF
--- a/Documentation/OrderedDictionary.md
+++ b/Documentation/OrderedDictionary.md
@@ -208,7 +208,7 @@ dictionary.
 The easiest way to achieve this is to make sure `Key` implements hashing
 following `Hashable`'s documented best practices. The conformance must
 implement the `hash(into:)` requirement, and every bit of information that
-is compared in `==` needs to be combined into the supplied `Hasher` value.
+is combined into the supplied `Hasher` value must be compared in `==`.
 When used correctly, `Hasher` produces high-quality, randomly seeded hash
 values that prevent repeatable hash collisions.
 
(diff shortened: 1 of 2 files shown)


xwu:
@hassila, the documentation is correct as-is but perhaps is easy to misread. The text you're proposing to edit isn't about _correctness_: it's in a section about _performance_ (literally, the subsection is titled "Performance").

As the preceding paragraph states, if all values of a type have the same hash despite being unequal, it's not a correctness issue but it _is_ a performance issue. Therefore, the paragraph that follows says that the _easiest_ (not _only_) correct way to achieve good performance is to follow best practices that include hashing everything that is compared in `==`. This helps the hasher to prevent repeatable hash collisions.

hassila:
Thanks @xwu, I did indeed misread the context and had recently read the discussion in the comment in the swift case I linked to in the description where they looked at this distinction in the standard library - now, there is obviously zero harm (and usually best practice) to include all bits in both the hashing and equality operations, but I thought it might makes sense to turn around the statement made so it still recommend this but also addresses the correctness (with no harm). 

If you disagree I'll just close this.

xwu:
Changing the statement as you propose, though, wouldn't give the correct advice about what one needs to do in order for the hasher to avoid collisions in order to maintain performance.

lorentey:
> now, there is obviously zero harm (and usually best practice) to include all bits in both the hashing and equality operations

This is an understatement. Doing this is not only _not_ harmful, but it is the only way to ensure that the documented performance promises of Set/Dictionary actually hold true.

A type that hashes less information than what it compares in `==` is, in almost all cases, an open invitation to reliability attacks.

(I added a similar note in the referenced issue, too. I'm a bit miffed that HashDoS vulnerabilities are still widely ignored/dismissed; hashing can be tricky enough without deliberately breaking it! 😛)

PR #298, opened 2023-06-22
Replace os() with canImport(Darwin)

brennobemoura:
With the release of Vision SDK, the refactor will need another `os(xrOS)` compiler directive conditional. A better and simple way to do this, as I saw in other projects, is to convert `os(iOS) || os(macOS) || os(tvOS) || os(watchOS)` to `canImport(Darwin)` leading to the same results.

### Checklist

### DIFF
--- a/Tests/_CollectionsTestSupport/AssertionContexts/CollectionTestCase.swift
+++ b/Tests/_CollectionsTestSupport/AssertionContexts/CollectionTestCase.swift
@@ -19,7 +19,7 @@ open class CollectionTestCase: XCTestCase {
 
   open var isAvailable: Bool { true }
 
-  #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+  #if canImport(Darwin)
   open override func invokeTest() {
     guard isAvailable else {
       print("\(Self.self) unavailable; skipping")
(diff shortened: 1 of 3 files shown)


lorentey:
The issue we are working around here only matters for macOS, iOS, watchOS and tvOS. It does not apply to visionOS. I do like the explicitness of listing the specific OSes directly; `canImport(Darwin)` is not an exact equivalent.

However, I think it would be fine to simply remove the `#if` statements here entirely -- the `if #available` check already ends up turning into a direct `return false` on platforms that aren't affected!

brennobemoura:
I agree with you. Based on what you’ve said, I'm started to understand that the problem here is not a `canImportant` solution case.

Tests/_CollectionsTestSupport/AssertionContexts/CollectionTestCase.swift
```diff
@@ -19,7 +19,7 @@ open class CollectionTestCase: XCTestCase {
 
   open var isAvailable: Bool { true }
 
-  #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+  #if canImport(Darwin)
```
lorentey, line 22: Ooh, this one is different. 

My worry is that `canImport(Darwin)` might return true if the project being built includes an unrelated module that coincidentally happens to be named `Darwin`. Targets not being cleanly isolated from each other has been a constant source of low-level issues.

I'd prefer if we did this:

```suggestion
  #if _runtime(_ObjC)
```

(Well, what I'd actually prefer is a Swift native test environment that understands availability, but this is a good stand-in. 😛)

brennobemoura: Awesome! I will close this PR, since this topic needs more clarification. The points you bring here is all valid to me. Now, I'm actually lost 🙈.

Thanks for sharing this!

PR #307, opened 2023-08-03
ART: Adaptive Radix Tree implementation

vishesh:
This PR implements Adaptive Radix Tree (ART)[1], a compact key-value search tree. It maps byte-sequences to arbitrary values while maintaining data in sorted order. ART can compress multiple nodes into single node and adaptively expand or shrink its internal nodes to avoid worse-case space consumption.

Within the broader swift-collection project, it can be considered a compact variant of SortedSet/SortedDictionary but at the cost of keys being convertible to byte-sequences.

[1] https://ieeexplore.ieee.org/abstract/document/6544812

### Checklist
- [ ] I've read the [Contribution Guidelines](/README.md#contributing-to-swift-collections)
- [ ] My contributions are licensed under the [Swift license](/LICENSE.txt).


ktoso (commented):
I'll be honest that this is a bit outside of the area of my expertise and I wasn't able to provide a lot of valuable input -- I know Karoy has some specific tricks to pull for such collections. However, all those hints are about to become invalidated by the arrival of moveonly types. At the same time, the allocation semantics this tree has I understand are something we _specifically_ want here -- so it is somewhat novel and not much precedent for it.

One thought is to maybe consider using the [benchmark package](https://github.com/ordo-one/package-benchmark) and measure the ARC traffic before we start doing these Unmanaged dances -- it really seems a bit iffy -- also because the manual retain/releases done on the class references are on a _strong_ class reference (`buf`) to a ManagedBuffer class so that would incur ARC trafi anyway -- probably best to kick off the benchmark package mentioned inline and measure ARC traffic before optimizing. 

We discussed with @lorentey and the collection should probably incubate for now outside of swift-collections but we'd like to try to help out polish it up -- even if for a specific use case; which may be easier than bringing it up to the long-term compatibility standards swift-collections would require...

I think that perhaps once we put the collection through some usage that'd be a good driver for brushing it up?

Thanks for the awesome work here so far @vishesh 🙏

ktoso:
I discussed the refernce counting a bit with @lorentey and here's some notes:

> The preferred setup is to either use strong references and automatic retain/release, or to keep child pointers represented with Unmanaged and use _withUnsafeGuaranteedRef and explicit retain/release operations.
> 
> Both options have significant drawbacks, but either option is better than trying to do an ad-hoc mix of the two.
>
> FWIW, in the prefix tree implementation of HashTreeCollections, the tree representation uses regular strong references to store the children of each node. There are no explicit retain/release operations anywhere in the code base — everything relies on regular strong reference semantics. (It tries to carefully avoid making any unnecessary copies, to avoid retain/release traffic. For example, child references aren’t loaded into local variables; they are accessed directly via storage pointers.)
> 
> (In the CHAMP data structure, child references and leaf items are mixed into the same storage buffer, and that heavily obscures the actual layout.)
>
> That module does also define a separate internal type that wraps an Unmanaged node reference. However, that has very limited functionality — it is primarily designed for use within the Index representation, and it requires extraordinary care to avoid accidentally dereferencing a stale value.

So it really feels like we should try to use a class to do the refcounting and use the package-benchmark to measure and optimize 🤔

Issue #308, opened 2023-08-09
`Deque` lacks `capacity`


glbrntt:
`Deque` is often used as a buffer. However, you can't always control how much data you accept. In these cases it usually makes sense to reclaim storage space if the capacity grows beyond some limit to avoid holding on to too much memory unnecessarily.

This isn't possible with `Deque` because its capacity is as an implementation detail. 

In [this thread](https://forums.swift.org/t/rfc-deque-should-it-have-a-capacity/47204) @lorentey suggested letting `Deque` shrink on removal along and adding `init(minimumCapacity:persistent:)`.  I think this would be sufficient although as described it has a minor drawback that it wouldn't lazily grow up to a capacity limit.

- I assume adding `capacity` is a non-starter based on [Deque.md](https://github.com/apple/swift-collections/blob/main/Documentation/Deque.md)?
- Has any more thought gone into allowing `Deque` to shrink automatically on removal?

lorentey:
Yes, I very intentionally omitted `Deque.capacity`, until I see a valid use case of it.

All code I've seen that tries to make decisions about shrinking storage based on `capacity` is overly sensitive to malloc behavior. `Deque` is using `malloc_size` to make use of every byte that it managed to allocate -- and if it happens to be given enough "free" space to put its capacity beyond some shrinking threshold naively implemented by Deque's client, then every operation may trigger the shrinkage, which would be quite bad.

The example code in the discussion you linked to was especially alarming to me:

```
self.buffers.removeAll(keepingCapacity: self.buffers.capacity < 16) // don't grow too much
```

16 seemed like a rather low value -- depending on the Element type and the whims of the system allocator, malloc may sometimes give us 16 items' worth of bytes even if we only ask for just a couple of them.

Given this, I'd prefer to figure out a more direct way to detect and recover from temporary allocation spikes. 

I generally dislike the idea of having the core resizing logic live in client code -- it ought to be part of the container implementation, as it needs to evolve with it.  Making Deque automatically shrink itself is one way to achieve this; another (possibly less disruptive) idea would be to expose an explicit operation to shrink the container to be near a certain target size. 

To do this, I think `Deque` would need to either give up on using malloc_size, or it would need to keep track of how much extra space malloc gave us.

glbrntt:
> Yes, I very intentionally omitted `Deque.capacity`, until I see a valid use case of it.
> 
[...]

PR #316, opened 2023-10-06
Let deque have its capacity shrunk

glbrntt:
### Description

Deque is often used as a buffer and buffers can be subject to "spiky" sizes; many elements may be buffered before being removed which can result in applications holding on to more memory than desired. Since the capacity of deque is an implementation detail, one way to reduce the size of the deque is asking it to shrink itself. 

Related discussion https://github.com/apple/swift-collections/issues/308

### Detailed Design

```swift
extension Deque {
  /// Reduces the capacity to store at least the specified number of elements.
  ///
  /// Use this method to request that the deque reduces its storage capacity.
  /// If the capacity of the deque is larger than the target capacity then
  /// the capacity _may_ be reduced. If the capacity of the deque is smaller
  /// than the requested capacity, or the deque contains more elements than
  /// the requested capacity, then the capacity won't be changed.
  ///
  /// - Parameters:
  ///   - targetCapacity: The requested capacity of the deque.
  ///
  /// - Complexity: O(`count`)
  public mutating func shrinkCapacity(_ targetCapacity: Int)
}
```

### Documentation

> How has the new feature been documented? 

On the new method only.

> Have the relevant portions of the guides in the Documentation folder been updated in addition to symbol-level documentation?

 Yes.

### Testing

> How is the new feature tested?

- By shrinking the capacity of deques of differing capacities
- By attempting to shrink a deque to a capacity smaller than the required capacity
- By attempting to shrink a deque to a capacity greater than its current capacity

### Performance

> How did you verify the new feature performs as expected?

I didn't.

### Source Impact

API addition only

### Checklist

### DIFF
--- a/Sources/DequeModule/Deque+Testing.swift
+++ b/Sources/DequeModule/Deque+Testing.swift
@@ -41,6 +41,11 @@ extension Deque {
     _storage.capacity
   }
 
+  @_spi(Testing)
+  public var _requestedCapacity: Int {
+    _storage.requestedCapacity
+  }
+
   /// The number of the storage slot in this deque that holds the first element.
   /// (Or would hold it after an insertion in case the deque is currently
   /// empty.)
@@ -67,7 +72,10 @@ extension Deque {
     precondition(contents.count <= capacity)
     let startSlot = _Slot(at: startSlot)
     let buffer = _DequeBuffer<Element>.create(minimumCapacity: capacity) { _ in
-      _DequeBufferHeader(capacity: capacity, count: contents.count, startSlot: startSlot)
+      _DequeBufferHeader(capacity: capacity, 
+                         requestedCapacity: capacity,
+                         count: contents.count,
+                         startSlot: startSlot)
     }
     let storage = Deque<Element>._Storage(unsafeDowncast(buffer, to: _DequeBuffer.self))
     if contents.count > 0 {
--- a/Tests/DequeTests/DequeTests.swift
+++ b/Tests/DequeTests/DequeTests.swift
@@ -374,5 +374,40 @@ final class DequeTests: CollectionTestCase {
     }
   }
 
+  func test_shrinkCapacityWhenEmpty() {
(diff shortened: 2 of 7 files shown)


hassila:
Why do you want separate control of shrinkage vs capacity reservation? Would it make sense to have reserveCapacity shrink if the current buffer is larger than the desired target capacity?

Tests/DequeTests/DequeTests.swift
```diff
@@ -374,5 +374,40 @@ final class DequeTests: CollectionTestCase {
     }
   }
 
+  func test_shrinkCapacityWhenEmpty() {
+    withEvery("capacity", in: [0, 1, 2, 5, 10, 50, 100]) { capacity in
+      var deque = Deque<Int>(minimumCapacity: capacity)
+
+      XCTAssertGreaterThanOrEqual(deque._capacity, capacity)
```
lorentey, line 381: Nit: Please use the custom `expect*` assertions from _CollectionsTestSupport -- `XCTAssert*` does not know about the context, so its failure messages will not identify the case that failed, making debugging much harder.

lorentey (commented):
Thanks for tackling this! 👍

Sources/DequeModule/Deque+Testing.swift
```diff
@@ -67,7 +72,10 @@ extension Deque {
     precondition(contents.count <= capacity)
     let startSlot = _Slot(at: startSlot)
     let buffer = _DequeBuffer<Element>.create(minimumCapacity: capacity) { _ in
-      _DequeBufferHeader(capacity: capacity, count: contents.count, startSlot: startSlot)
+      _DequeBufferHeader(capacity: capacity, 
+                         requestedCapacity: capacity,
+                         count: contents.count,
+                         startSlot: startSlot)
```
aluco100:
```suggestion
      _DequeBufferHeader(
          capacity: capacity, 
           requestedCapacity: capacity,
           count: contents.count,
           startSlot: startSlot)
```

PR #330, opened 2023-11-24
build: support using swift-collections in larger projects

compnerd:
This enables the use of swift-collections with CMake in a larger project with CMake. In particular, this enables the use of swift-collections with FetchContent to enable integration with other projects for vendoring.


compnerd:
@swift-ci please test

lorentey (approved):

Issue #334, opened 2023-12-05
OrderedDictionary.values and OrderedSet.values Performance Problems when SwiftUI.List is paired with SwiftUI.NavigationStack


vanvoorden:
https://developer.apple.com/forums/thread/741484

Hi! I'm copying this information over from Apple DevForums. My guess here is that there is a potential optimization inside the SwiftUI library itself that can clean up these perf issues… but I'll go ahead and post the details here just in case there is something else inside the `OrderedDictionary` itself that should be optimized. Thanks!

### Information

- **Package version:** 1.0.5
- **Platform version:** 17.0 Simulator and Device
- **Swift version:** Apple Swift version 5.9 (swiftlang-5.9.0.128.108 clang-1500.0.40.1)

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] I've searched for [existing GitHub issues](https://github.com/apple/swift-collections/issues).

### Steps to Reproduce

https://gist.github.com/vanvoorden/1c7c6ed08898de7f4b8619147537c0eb

Hi! Has anyone here seen weird performance problems when passing values from a `Collections.OrderedDictionary` to a `SwiftUI.List` inside `SwiftUI.NavigationStack`? I see some performance problems that go away when switching back to `SwiftUI.NavigationView`. At this point, I'm not sure if it's a problem in the `OrderedDictionary` implementation or not. Anyone want to take a look and see if you can help track this down?

For a simple backing store to drive a simple SwiftUI app displaying a `List` of data, the `OrderedDictionary` (from `swift-collections`) has the advantage of giving us the option to access (by an arbitrary `key`) in constant time. The `OrderedDictionary` exposes a `values` property that conforms to `RandomAccessCollection`. Since we can pass a `RandomAccessCollection` as data to generate a `List`, we should hope to be able to pass the `OrderedDictionary.values` the same way we would pass an `Array`.

This is leading to some strange performance problems that seem to happen when our `List` is contained in a `NavigationStack` (and we push and pop a detail view on the stack). Here is a simple example:

```
import Collections
import SwiftUI

let NavigationDemoRange = 0...999999

extension Int: Identifiable {
  public var id: Self {
    return self
  }
}

@main struct NavigationDemoApp: App {
  let dictionary = OrderedDictionary(
    uniqueKeys: NavigationDemoRange,
    values: NavigationDemoRange
  )
}

extension NavigationDemoApp {
  var body: some Scene {
    WindowGroup {
[...]

PR #335, opened 2023-12-08
[OrderedDictionary] forward ordered dictionary values equality to values property

vanvoorden:
### Background
https://github.com/apple/swift-collections/issues/334

The current implementation of `OrderedDictionary.Values.==` forwards to `Sequence.elementsEqual`, which needs to perform a linear-time (`O[N]`) comparison over both collections.[1]

Since the "backing store" of our `OrderedDictionary.Values` is a `ContiguousArray`, we can forward that equality check through to `ContiguousArray`. If both `Values` instances point to the same identity, the `ContiguousArray` implementation will return early from the comparison.[2]

We perform a similar check (against the `ContiguousArray`) in our implementation of `OrderedDictionary.==`.[3]

[1] https://github.com/apple/swift/blob/main/stdlib/public/core/SequenceAlgorithms.swift#L319-L336
[2] https://github.com/apple/swift/blob/main/stdlib/public/core/ContiguousArray.swift#L1318-L1321
[3] https://github.com/apple/swift-collections/blob/main/Sources/OrderedCollections/OrderedDictionary/OrderedDictionary%2BEquatable.swift#L20-L22

---

### Changes

We migrate from:
```
  left.elementsEqual(right)
}
```
We migrate to:
```
  left._base._values == right._base._values
}
```

---

### Test Plan

Three new tests are added:

* `OrderedDictionaryValueTests.test_values_getter_equal`
* `OrderedDictionaryValueTests.test_values_getter_not_equal`
* `OrderedDictionaryValueTests.test_values_getter_not_equal_elements`
---

### Next Steps

https://github.com/apple/swift-collections/blob/main/Sources/OrderedCollections/OrderedSet/OrderedSet%2BEquatable.swift#L25-L27

The `OrderedSet` implements similar equality logic. We can implement a similar diff and forward the equality check to the `ContiguousArray` backing store. We can either add that work here in this diff or create a separate diff.

---

### Checklist

### DIFF
--- a/Sources/_CollectionsTestSupport/AssertionContexts/Assertions.swift
+++ b/Sources/_CollectionsTestSupport/AssertionContexts/Assertions.swift
@@ -300,22 +300,6 @@ public func expectEqualElements<S1: Sequence, S2: Sequence>(
     message, trapping: trapping, file: file, line: line)
 }
 
-public func expectNotEqualElements<S1: Sequence, S2: Sequence>(
-  _ left: S1,
-  _ right: S2,
-  _ message: @autoclosure () -> String = "",
-  trapping: Bool = false,
-  file: StaticString = #file,
-  line: UInt = #line
-) where S1.Element == S2.Element, S1.Element: Equatable {
-  let left = Array(left)
-  let right = Array(right)
-  if !left.elementsEqual(right) { return }
-  _expectFailure(
-    "'\(left)' does have equal elements to '\(right)'",
-    message, trapping: trapping, file: file, line: line)
-}
-
 public func expectEquivalentElements<S1: Sequence, S2: Sequence>(
   _ left: S1,
   _ right: S2,
--- a/Sources/_CollectionsTestSupport/AssertionContexts/Assertions.swift
+++ b/Sources/_CollectionsTestSupport/AssertionContexts/Assertions.swift
@@ -300,6 +300,22 @@ public func expectEqualElements<S1: Sequence, S2: Sequence>(
     message, trapping: trapping, file: file, line: line)
 }
 
+public func expectNotEqualElements<S1: Sequence, S2: Sequence>(
+  _ left: S1,
+  _ right: S2,
+  _ message: @autoclosure () -> String = "",
+  trapping: Bool = false,
+  file: StaticString = #file,
+  line: UInt = #line
+) where S1.Element == S2.Element, S1.Element: Equatable {
+  let left = Array(left)
+  let right = Array(right)
+  if !left.elementsEqual(right) { return }
+  _expectFailure(
+    "'\(left)' does have equal elements to '\(right)'",
+    message, trapping: trapping, file: file, line: line)
+}
+
 public func expectEquivalentElements<S1: Sequence, S2: Sequence>(
   _ left: S1,
   _ right: S2,
--- a/Tests/OrderedCollectionsTests/OrderedDictionary/OrderedDictionary+Values Tests.swift
+++ b/Tests/OrderedCollectionsTests/OrderedDictionary/OrderedDictionary+Values Tests.swift
@@ -28,7 +28,8 @@ class OrderedDictionaryValueTests: CollectionTestCase {
       "three": 3,
       "four": 4,
     ]
-    expectEqual(left.values, right.values)
+    expectEqual(left.values, left.values) // Identity fast path
+    expectEqual(left.values, right.values) // Linear algorithm
   }
   
   func test_values_getter_not_equal() {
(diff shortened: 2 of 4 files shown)


vanvoorden:
@swift-ci test

vanvoorden:
@ingoem thanks for suggesting the reference equality check!

Tests/_CollectionsTestSupport/AssertionContexts/Assertions.swift
```diff
@@ -300,6 +300,22 @@ public func expectEqualElements<S1: Sequence, S2: Sequence>(
     message, trapping: trapping, file: file, line: line)
 }
 
+public func expectNotEqualElements<S1: Sequence, S2: Sequence>(
```
lorentey, line 303: Huh, perhaps we should have some sort of DSL for these expectations. But this is fine!

lorentey (approved):
Nice! 👍

vanvoorden:
> I think it would make sense to ship this in the next patch release, rather than having it linger on main indefinitely.

@lorentey Sounds good. Thanks!

vanvoorden:
@lorentey Is there a CI job that can run `Benchmarks` on a diff to detect for regressions or improvements against `base`?

vanvoorden:
Here are benchmarks running locally. The reason that `OrderedDictionary.==` is scaling linearly is because we are waiting on https://github.com/apple/swift-collections/pull/340. Optimizing the `OrderedSet` equality will optimize this one.

Tests/OrderedCollectionsTests/OrderedDictionary/OrderedDictionary+Values Tests.swift
```diff
@@ -15,7 +15,40 @@ import XCTest
...
+      "three": 3,
+      "four": 4,
+    ]
+    let right: OrderedDictionary = [
+      "one": 1,
+      "two": 2,
+      "three": 3,
+      "four": 4,
+    ]
+    expectEqual(left.values, right.values)
```
lorentey, line 31: We should also make sure to exercise the new code path:

```suggestion
    expectEqual(left.values, left.values) // Identity fast path
    expectEqual(left.values, right.values) // Linear algorithm
```

Sources/_CollectionsTestSupport/AssertionContexts/Assertions.swift
```diff
@@ -300,6 +300,22 @@ public func expectEqualElements<S1: Sequence, S2: Sequence>(
...
+  line: UInt = #line
+) where S1.Element == S2.Element, S1.Element: Equatable {
+  let left = Array(left)
+  let right = Array(right)
+  if !left.elementsEqual(right) { return }
+  _expectFailure(
+    "'\(left)' does have equal elements to '\(right)'",
+    message, trapping: trapping, file: file, line: line)
+}
+
```
lorentey, line 318: On second look, I recommend getting rid of this function -- it doesn't carry its weight.

```suggestion
```

(Also see the comment on `test_values_getter_not_equal_elements`.)

Tests/OrderedCollectionsTests/OrderedDictionary/OrderedDictionary+Values Tests.swift
```diff
@@ -24,6 +57,16 @@ class OrderedDictionaryValueTests: CollectionTestCase {
...
+  
+  func test_values_getter_not_equal_elements() {
+    let d: OrderedDictionary = [
+      "one": 1,
+      "two": 2,
+      "three": 3,
+      "four": 4,
+    ]
+    expectNotEqualElements(d.values, [1, 2, 3, 4, 5])
+  }
```
lorentey, line 69: On second look, this test seems to be mostly testing the standard `elementsEqual` algorithm; it doesn't add anything that's specific to `OrderedDictionary.Values` but that isn't already covered by the `checkBidirectionalCollection(d.values, expectedContents: values)` test in an earlier test method above.

That protocol validation test verifies that the sequence has the expected contents (in addition to a bunch of other things), which is all we need! We do not need a specific test to verify that the contents do not also match some other list -- that can be safely assumed to be the case. 

(Such a check may be useful if we were testing `elementsEqual` itself, but this isn't the right place for that.)

```suggestion
```

lorentey (approved):
Thanks for rebasing. Nice work on the benchmarks! 👍

Reviewing #340 made me realize what's been bothering me about `expectNotEqualElements`; I added new recommendations to remove it and the test that exercises it. (Please see the notes below.)

vanvoorden:
@lorentey Thanks for reviewing!

lorentey:
@swift-ci test

lorentey:
(AFAIK, CI runs are still expected to fail on macOS. This is "fine", and it won't block us from landing this.)

PR #338, opened 2023-12-08
build: tweak search path for embedding

compnerd:
Correct the path to allow using swift-collections embedded in a larger project.  This is meant to assist with building LSP with vendored dependencies.


compnerd:
CC: @lorentey

lorentey (approved):

PR #339, opened 2023-12-08
build: tweak search path for embedding

compnerd:
Correct the path to allow using swift-collections embedded in a larger project.  This is meant to assist with building LSP with vendored dependencies.


lorentey:
We don't need to duplicate PRs like this and #338, unless there are material changes between them -- changes flow from upstream to downstream dev branches (`release/1.0` → `release/1.1` → `main`). (Unfortunately we do not have automergers set up, so this is currently done manually, as needed.)

PR #340, opened 2023-12-09
[OrderedSet] forward ordered set equality to elements property

vanvoorden:
### Background
https://github.com/apple/swift-collections/pull/335

Similar to `OrderedDictionary`, we can forward our `OrderedSet` equality checks down to the `ContiguousArray` instance to take advantage of early return when both arrays point to the same identity (no need to perform a linear comparison over all elements).

A side-effect of this optimization is we also optimize `OrderedDictionary.==`, which performs an equality check on the `keys` property (which is an `OrderedSet`).[1]

[1] https://github.com/apple/swift-collections/blob/main/Sources/OrderedCollections/OrderedDictionary/OrderedDictionary%2BEquatable.swift#L20-L22

---

### Changes

From:
```
  left.elementsEqual(right)
}
```
To:
```
  left._elements == right._elements
}
```

We can also migrate `OrderedSet.SubSequence`. From:
```
  left.elementsEqual(right)
}
```
To:
```
  left._base._elements[left._bounds] == right._base._elements[right._bounds]
}
```

---

### Test Plan

Five new tests are added:

- `OrderedSetTests.test_equal`
- `OrderedSetTests.test_not_equal`
- `OrderedSetTests.test_equal_elements`
- `OrderedSetTests.test_subsequence_equality`
- `OrderedSetTests.test_subsequence_not_equality`

---

### Checklist

### DIFF
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+SubSequence.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+SubSequence.swift
@@ -340,7 +340,7 @@ extension OrderedSet.SubSequence: RandomAccessCollection {
 extension OrderedSet.SubSequence: Equatable {
   @inlinable
   public static func ==(left: Self, right: Self) -> Bool {
-    left.elementsEqual(right)
+    left._base._elements == right._base._elements
   }
 }
 
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+SubSequence.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+SubSequence.swift
@@ -340,7 +340,7 @@ extension OrderedSet.SubSequence: RandomAccessCollection {
 extension OrderedSet.SubSequence: Equatable {
   @inlinable
   public static func ==(left: Self, right: Self) -> Bool {
-    left._base._elements[_bounds] == right._base._elements[_bounds]
+    left._base._elements[left._bounds] == right._base._elements[right._bounds]
   }
 }
 
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+SubSequence.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+SubSequence.swift
@@ -340,7 +340,7 @@ extension OrderedSet.SubSequence: RandomAccessCollection {
 extension OrderedSet.SubSequence: Equatable {
   @inlinable
   public static func ==(left: Self, right: Self) -> Bool {
-    left._base._elements == right._base._elements
+    left._base._elements[_bounds] == right._base._elements[_bounds]
   }
 }
 
--- a/Tests/OrderedCollectionsTests/OrderedSet/OrderedSetTests.swift
+++ b/Tests/OrderedCollectionsTests/OrderedSet/OrderedSetTests.swift
@@ -1360,4 +1360,52 @@ class OrderedSetTests: CollectionTestCase {
       }
     }
   }
+  
+  func test_equal() {
+    withEvery("count", in: 0 ..< 20) { count in
+      let set = OrderedSet(0 ..< count)
+      let copy = set
+      expectEqual(copy, set)
+    }
+  }
+  
+  func test_not_equal() {
+    withEvery("count", in: 0 ..< 20) { count in
+      let left = OrderedSet(0 ..< count)
+      let right = OrderedSet(0 ... count)
+      expectNotEqual(left, right)
+    }
+  }
+  
+  func test_equal_elements() {
+    withEvery("count", in: 0 ..< 20) { count in
+      let set = OrderedSet(0 ..< count)
+      let copy = set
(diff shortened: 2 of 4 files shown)


vanvoorden:
Here's what the benchmarks look like running locally.

vanvoorden:
Here are more benchmarks running locally.

Tests/OrderedCollectionsTests/OrderedSet/OrderedSetTests.swift
```diff
@@ -1360,4 +1360,50 @@ class OrderedSetTests: CollectionTestCase {
...
+      expectNotEqual(left, right)
+    }
+  }
+  
+  func test_subsequence_equal_elements() {
+    withEvery("count", in: 0 ..< 20) { count in
+      let subsequence = OrderedSet(0 ..< count)[0 ..< count]
+      expectEqualElements(subsequence, 0 ..< count)
+    }
+  }
```
lorentey, line 1408: Hm, these seem little different to simple tests such as this:

```swift
let a: OrderedSet = [0, 1, 2, 3, 4]
expectEqual(a[1 ..< 3], a[1 ..< 3])
```

How about something like this?

```suggestion
  func test_subsequence_equality() {
    let c = 5
    let items1 = OrderedSet(0 ..< c)
    let items2 = OrderedSet(0 ..< c)
    withEvery("i", in: 0 ... c) { i in
      withEvery("j", in: i ... c) { j in
        expectEqual(items1[i ..< j], items1[i ..< j]) // Reflective fast path based on identity
        expectEqual(items1[i ..< j], items2[i ..< j]) // Linear path
        var items3 = OrderedSet(i ..< j)
        items3.remove(j)
        items3.insert(j + 1)
        expectNotEqual(items1[i ..< j], items3[...]) // Slow negative path
      }
    }
  }
```

Note that the `expectEqualElements` case is already exhaustively covered by `checkBidirectionalCollection` in an earlier test.

lorentey (changes requested):
Great catch again -- we'll definitely want to have this change. 👍 

I think I found a correctness issue, though; it also indicates that we should do another pass on the test additions, to improve coverage. 🤔 I commented some suggested changes. (Beware that I haven't tried them, though. 😛)

Sources/OrderedCollections/OrderedSet/OrderedSet+SubSequence.swift
```diff
@@ -340,7 +340,7 @@ extension OrderedSet.SubSequence: RandomAccessCollection {
 extension OrderedSet.SubSequence: Equatable {
   @inlinable
   public static func ==(left: Self, right: Self) -> Bool {
-    left.elementsEqual(right)
+    left._base._elements == right._base._elements
```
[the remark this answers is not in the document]
vanvoorden: > This doesn't look right -- comparing _base._elements would compare the entire contents of the base sets of both inputs, rather than just the actual slices that they represent.

Ahh… that was a bad bug! Thanks!

Tests/OrderedCollectionsTests/OrderedSet/OrderedSetTests.swift
```diff
@@ -1360,4 +1360,56 @@ class OrderedSetTests: CollectionTestCase {
...
+      expectNotEqual(items1[0 ..< c], leftSlice)  //  same identity
+      expectNotEqual(items2[0 ..< c], leftSlice)  //  different identity
+      
+      let rightSlice = items1[i + 1 ..< c]
+      expectNotEqual(items1[0 ..< c], rightSlice) //  same identity
+      expectNotEqual(items2[0 ..< c], rightSlice) //  different identity
+      withEvery("j", in: i ... c) { j in
+      }
+    }
+  }
```
vanvoorden, line 1414: @lorentey I confirmed that this test "broke" on the previous equality check with the bug.

Tests/OrderedCollectionsTests/OrderedSet/OrderedSetTests.swift
```diff
@@ -1360,4 +1360,54 @@ class OrderedSetTests: CollectionTestCase {
...
+    withEvery("i", in: 0 ..< c) { i in
+      let leftSlice = items1[0 ..< i]
+      expectNotEqual(items1[0 ..< c], leftSlice)  //  same identity
+      expectNotEqual(items2[0 ..< c], leftSlice)  //  different identity
+      
+      let rightSlice = items1[i + 1 ..< c]
+      expectNotEqual(items1[0 ..< c], rightSlice) //  same identity
+      expectNotEqual(items2[0 ..< c], rightSlice) //  different identity
+    }
+  }
```
vanvoorden, line 1412: @lorentey I confirmed that this test "broke" on the previous equality check with the bug.

lorentey (approved):
This looks ready! 👍

lorentey:
@swift-ci test

PR #341, opened 2023-12-12
[Deque] check deque equality with buffer identity

vanvoorden:
### Background
https://github.com/apple/swift-collections/pull/335

Similar to what we added for `OrderedDictionary`, we can try and optimize our `Deque` equality checks for the case where two `Deque` instances point to the same identity.

Unlike `OrderedDictionary`, our `Deque` is not implemented on `ContiguousArray`. We can try and mimic the approach from `ContiguousArray` here in `Deque` by checking against the identity of the underlying `ManagedBufferPointer` instance.

---

### Changes

From:
```
  left.elementsEqual(right)
}
```
To:
```
let lhsCount = left.count
if lhsCount != right.count {
  return false
}

// Test referential equality.
if lhsCount == 0 || left._storage.isIdentical(to: right._storage) {
  return true
}

return left.elementsEqual(right)
```

---

### Test Plan

We leverage the existing `DequeTests`.

---

### Open Questions

Are there any additional edge-casey behaviors we should be testing for?

---

### Checklist

### DIFF
--- a/Sources/DequeModule/Deque+Equatable.swift
+++ b/Sources/DequeModule/Deque+Equatable.swift
@@ -17,6 +17,16 @@ extension Deque: Equatable where Element: Equatable {
   /// - Complexity: O(`min(left.count, right.count)`)
   @inlinable
   public static func ==(left: Self, right: Self) -> Bool {
+    let lhsCount = left.count
+    if lhsCount != right.count {
+      return false
+    }
+
+    // Test referential equality.
+    if lhsCount == 0 || left._storage._buffer.buffer === right._storage._buffer.buffer {
+      return true
+    }
+    
     return left.elementsEqual(right)
   }
 }
--- a/Sources/DequeModule/Deque+Equatable.swift
+++ b/Sources/DequeModule/Deque+Equatable.swift
@@ -23,7 +23,7 @@ extension Deque: Equatable where Element: Equatable {
     }
 
     // Test referential equality.
-    if lhsCount == 0 || left._storage._buffer.buffer === right._storage._buffer.buffer {
+    if lhsCount == 0 || left._storage.isIdentical(to: right._storage) {
       return true
     }
     
--- a/Sources/DequeModule/Deque._Storage.swift
+++ b/Sources/DequeModule/Deque._Storage.swift
@@ -212,3 +212,11 @@ extension Deque._Storage {
     }
   }
 }
+
+extension Deque._Storage {
+  @inlinable
+  @inline(__always)
+  internal func isIdentical(to other: Self) -> Bool {
+    self._buffer.buffer === other._buffer.buffer
+  }
+}
--- a/Tests/DequeTests/DequeTests.swift
+++ b/Tests/DequeTests/DequeTests.swift
@@ -369,6 +369,23 @@ final class DequeTests: CollectionTestCase {
       }
     }
   }
+  
+  func test_equal() {
+    let deque: Deque<Int> = [1, 2, 3, 4]
+    let copy = deque
+    expectEqual(deque, copy)
+  }
+  
+  func test_not_equal() {
+    let left: Deque<Int> = [1, 2, 3, 4]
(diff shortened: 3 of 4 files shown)


vanvoorden:
Here are benchmarks from running locally.

Tests/DequeTests/DequeTests.swift
```diff
@@ -369,6 +369,23 @@ final class DequeTests: CollectionTestCase {
...
+    let deque: Deque<Int> = [1, 2, 3, 4]
+    let copy = deque
+    expectEqual(deque, copy)
+  }
+  
+  func test_not_equal() {
+    let left: Deque<Int> = [1, 2, 3, 4]
+    let right: Deque<Int> = [1, 2, 3, 4, 5]
+    expectNotEqual(left, right)
+  }
```
lorentey, line 383: The `==` implementation is already covered in a more systematic way in the `test_Equatable_Hashable` test above. I think we can safely remove these.

```suggestion
```

Tests/DequeTests/DequeTests.swift
```diff
@@ -369,6 +369,23 @@ final class DequeTests: CollectionTestCase {
...
+  
+  func test_equal_elements() {
+    let deque: Deque<Int> = [1, 2, 3, 4]
+    expectEqualElements(deque, [1, 2, 3, 4])
+  }
```
lorentey, line 388: Similarly, the `test_CollectionConformance` test is already systematically checking that deques (of various shapes) do always produce the expected contents. (It also checks all possible methods of iterating through the deque, not just the one that `elementsEqual` uses.) This test can therefore be safely omitted.

```suggestion
```

Sources/DequeModule/Deque+Equatable.swift
```diff
@@ -17,6 +17,16 @@ extension Deque: Equatable where Element: Equatable {
   /// - Complexity: O(`min(left.count, right.count)`)
   @inlinable
   public static func ==(left: Self, right: Self) -> Bool {
+    let lhsCount = left.count
+    if lhsCount != right.count {
+      return false
+    }
+
+    // Test referential equality.
+    if lhsCount == 0 || left._storage._buffer.buffer === right._storage._buffer.buffer {
```
lorentey, line 26: Super minor nit: We usually try to avoid directly accessing low-level layers within the implementation of high-level public API like this `==` implementation. It could be worth adding an `isIdentical(to:)` method to `Deque._Storage`, and then calling it here, instead of directly extracting the buffer references.

lorentey (approved):
Nice 👍

I picked some minor nits below.

lorentey:
Note to self: letting `Deque.SubSequence` be the default `Slice<Deque>` means that it can't conform to `Equatable`/`Hashable`, and that is probably a defect in the Standard Library.

Sources/DequeModule/Deque._Storage.swift
```diff
@@ -212,3 +212,11 @@ extension Deque._Storage {
     }
   }
 }
+
+extension Deque._Storage {
+  @inlinable
+  @inline(__always)
```
vanvoorden, line 218: @lorentey Do these `inline` attributes look appropriate here?

lorentey: This one looks fine to me! This package generally wants all generic code to be `@inlinable` to allow specialization (and to prevent having to materialize metatypes just to call this function). (This may not remain a tenable approach for long, but as long as these constructs are distributed as a package, the drawbacks of exposing function bodies are less pressing than e.g. if they shipped in the stdlib.)

`@inline(__always)` encourages the optimizer to actually inline the body more often into clients. In the case of `==`, I expect this doesn't matter much either way.

lorentey:
@swift-ci test

lorentey (approved):
This looks ready to land! 🚢

vanvoorden:
@lorentey Thanks for reviewing!

PR #352, opened 2024-01-28
[TreeDictionary][Keys] Add Equatable and Hashable Conformance to TreeDictionary.Keys 

vanvoorden:
# Background

Unlike `Swift.Dictionary.Keys`[^1], `TreeDictionary.Keys` does not currently conform to `Equatable`. Achieving more parity with the standard library dictionary could lower friction for more engineers to migrate to `TreeDictionary` in their own repos.

# Changes

We can copy the approach from `TreeSet`:

```
public static func == (left: Self, right: Self) -> Bool {
  left._base._root.isEqualSet(to: right._base._root, by: { _, _ in true })
}
```

Since we don't care if *values* are equal (only the keys), we can hardcode the equality closure to always return `true`.

Similar to our equality check on `TreeDictionary`, we expect to return `true` in constant-time if two values are identical by reference (no need for a linear-time check).

# Test Plan

We add `TreeDictionaryKeysTests.test_isEqual_exhaustive`. We can expand on these tests if necessary.

# Benchmarks

```
swift run -c release benchmark run results.json --cycles 5 --tasks "TreeDictionary<Int, Int> equality, unique" "TreeDictionary<Int, Int> equality, shared"

swift run -c release benchmark render results.json tree-dictionary.png --amortized false --tasks "TreeDictionary<Int, Int> equality, unique" "TreeDictionary<Int, Int> equality, shared"
```
tree-dictionary

```
swift run -c release benchmark run results.json --cycles 5 --tasks "TreeDictionary<Int, Int>.Keys equality, unique" "TreeDictionary<Int, Int>.Keys equality, shared"

swift run -c release benchmark render results.json tree-dictionary-keys.png --amortized false --tasks "TreeDictionary<Int, Int>.Keys equality, unique" "TreeDictionary<Int, Int>.Keys equality, shared"
```
tree-dictionary-keys

# Checklist

[^1]: https://developer.apple.com/documentation/swift/dictionary/keys-swift.struct


lorentey (approved):
Looks good! Thanks for catching this.

Generally, it's good practice for Equatable types to also conform to Hashable; I'll think we'll need that too.

vanvoorden:
> Generally, it's good practice for Equatable types to also conform to Mashable

https://developer.apple.com/documentation/swift/dictionary/keys-swift.struct#conforms-to

@lorentey Hmm… it looks like `Swift.Dictionary.Keys` actually *does not* conform to Hashable (but does conform to Equatable). There might be some interesting use cases enabled when an engineer can assume `TreeDictionary.Keys` was Hashable… but it looks like the precedent from the Standard Library `Dictionary` is that just Equatable might be good enough for now.

lorentey:
Unfortunately the Standard Library does not set a good example here, as we cannot easily add missing conformances that we forgot to ship in the initial release. (They are potentially ABI breaking.)

vanvoorden:
> we cannot easily add missing conformances there that we forgot to ship in the initial release

https://github.com/apple/swift-collections/blob/release/1.1/Sources/HashTreeCollections/TreeSet/TreeSet%2BHashable.swift

@lorentey Hmm… something like what we do for `TreeSet` might work here? Want me to give it a try and post a diff?

lorentey:
Yes, that's precisely the pattern! :+1

vanvoorden:
> Yes, that's precisely the pattern!

@lorentey Sounds good. I can try that. Would it be ok to add a new commit here to this diff for another review to save us the merge conflict from trying to land both?

lorentey:
For sure! Adding a commit here is best.

Tests/HashTreeCollectionsTests/TreeDictionary.Keys Tests.swift
```diff
@@ -105,5 +105,45 @@ class TreeDictionaryKeysTests: CollectionTestCase {
...
+      [[100: 100], [100: 100]],
+      [[1: 100, 2: 200], [2: 200, 1: 100]],
+      [[1: 100, 2: 200, 3: 300],
+       [1: 100, 3: 300, 2: 200],
+       [2: 200, 1: 100, 3: 300],
+       [2: 200, 3: 300, 1: 100],
+       [3: 300, 1: 100, 2: 200],
+       [3: 300, 2: 200, 1: 100]]
+    ]
+    checkHashable(equivalenceClasses: samples.map { $0.map { $0.keys }})
```
vanvoorden, line 146: https://gist.github.com/vanvoorden/5895caeef3417920e262e4c85edcbf0c

@lorentey I'm seeing a ton of test failures here… is there an obvious bug here you think I am missing?

vanvoorden:
```
diff --git a/Tests/HashTreeCollectionsTests/TreeSet Tests.swift b/Tests/HashTreeCollectionsTests/TreeSet Tests.swift
index aa3f675..b309f07 100644
--- a/Tests/HashTreeCollectionsTests/TreeSet Tests.swift	
+++ b/Tests/HashTreeCollectionsTests/TreeSet Tests.swift	
@@ -684,6 +684,9 @@ class TreeSetTests: CollectionTestCase {
       [
         ["a", "b"], ["b", "a"],
       ],
+      [
+        ["a", "b"], ["b", "a"],
+      ],
       [
         ["a", "d"], ["d", "a"],
       ],
```
@lorentey Trying to repeat a pair of sets in this test leads to a failure… is it possible that `checkHashable` helper is expected to fail when we duplicate members?

Tests/HashTreeCollectionsTests/TreeDictionary.Keys Tests.swift
```diff
@@ -105,5 +105,85 @@ class TreeDictionaryKeysTests: CollectionTestCase {
...
+        ["b", "a", "c"], ["b", "c", "a"],
+        ["c", "a", "b"], ["c", "b", "a"],
+      ],
+      [
+        ["a", "d", "e"], ["a", "e", "d"],
+        ["d", "a", "e"], ["d", "e", "a"],
+        ["e", "a", "d"], ["e", "d", "a"],
+      ],
+    ]
+    checkHashable(equivalenceClasses: classes.map { $0.map { TreeDictionary(keys: $0, valueGenerator: { $0 }).keys }})
```
vanvoorden, line 186: @lorentey This test is passing… does this look legit to you?

Tests/HashTreeCollectionsTests/TreeDictionary.Keys Tests.swift
lorentey: Oh, yes! That is precisely it: the test verifies not only that items in the same equivalence class compare equal, but also that items in other classes don't.

We'll also want the test to verify that `==` and `hash(into:)` ignore the values, so we should exercise those differences too. (I suppose these two cases do that, though!)

Here is one possible way to update things:

```
let samples: [[TreeDictionary<Int, Int>]] = [
       [[:], [:]],
       [[1: 100], [1: 100], [1: 1], [1: 2]],
       [[2: 200], [2: 201]],
       [[3: 300], [3: 301]],
       [[100: 1], [100: 1], [100: 100], [100: 100]],
       [[1: 100, 2: 200], [2: 200, 1: 100], [1: 100, 2: 201]],
       [[1: 100, 2: 200, 3: 300],
        [1: 100, 3: 300, 2: 200],
        [2: 200, 1: 100, 3: 300],
        [2: 200, 3: 300, 1: 100],
        [3: 300, 1: 100, 2: 200],
        [3: 300, 2: 200, 1: 100]]
     ]
```

(I haven't tried this, so beware of thinkos. 😅)

Tests/HashTreeCollectionsTests/TreeDictionary.Keys Tests.swift
```diff
@@ -105,5 +105,66 @@ class TreeDictionaryKeysTests: CollectionTestCase {
...
+        ["c", "a", "b"], ["c", "b", "a"],
+      ],
+      [
+        ["a", "d", "e"], ["a", "e", "d"],
+        ["d", "a", "e"], ["d", "e", "a"],
+        ["e", "a", "d"], ["e", "d", "a"],
+      ],
+    ]
+    let keys = strings.map { $0.map { TreeDictionary(uniqueKeysWithValues: $0.map { ($0, UUID().uuidString) }).keys }}
+    checkHashable(equivalenceClasses: keys)
```
vanvoorden, line 167: @lorentey How about this? I'm generating random values with `UUID().uuidString` so we can confirm the `keys` are still equal when values are not equal. I could try something different if we prefer for those values to be generated deterministically (the same every time).

lorentey: Nice! I was actually typing in a comment recommending something along these lines -- a deterministic alternative would be to use a sequential counter, but randomizing things works just as well.

Instead of Foundation's UUID, I think we should just use `SystemRandomNumberGenerator`.

lorentey (approved):
This looks ready to land! Thanks very much for taking this on. 👍

lorentey:
@swift-ci test linux platform

PR #388, opened 2024-06-12
[Future] Introduce the Inout type

Azoy:


Package.swift
```diff
@@ -52,7 +52,12 @@ var defines: [String] = [
 //  "COLLECTIONS_SINGLE_MODULE",
 ]
 
-var _settings: [SwiftSetting] = defines.map { .define($0) }
+nonisolated(unsafe)
```
glessard, line 55: What. Or just make it a `let` binding?

glessard (approved):
This is a fun prototype.

PR #392, opened 2024-06-18
[CMake] Reduce path lengths in single-module build

jmschonfeld:
Previously, the single module build would add each modules sources to a different list, and then we would combine the list prepending each element with a path like `../BitArray` or `../RopeModule`. This resulted in extremely long build paths when built as a CMake FetchContent dependency which exceeded the maximum path length on Windows. To resolve this issue, we now create the single module target at the `Sources` level, and then call `target_sources` in each subdirectory with a normal relative path such as `BitArray+Description.swift`. This results in significantly smaller build directory structures keeping the path length within the Windows maximum and also cleans up the CMake a bit.

### Checklist


lorentey (approved):
Huh! 👍

PR #395, opened 2024-06-25
Reduce the size of the _FoundationCollections toolchain module

jmschonfeld:
The number of object files in the single _FoundationCollections toolchain module was causing issues on Windows by surpassing the maximum command line invocation length. To work around this for now, we reduce the size of the module by not including certain sub-modules. This will unblock the swift-foundation re-core work. At a later time, we could come back to adjust the CMake build to use response files to avoid a large command line invocation if necessary as a full solution.

### Checklist


lorentey (approved):
Reluctant thumbs up! Did enabling response files not work right?

jmschonfeld:
Spoke offline but unfortunately this was the simplest solution that unblocks the swift-foundation work at the moment, I wasn't able to immediately get the response files working but I'll followup to see if we can update to that approach in a later PR

PR #401, opened 2024-07-04
Changed 'count == 0' to 'isEmpty' and add 'final' keyword to class

quokkaKyu:
Hello, when checking whether the value of Collection Types is empty, how about using 'isEmpty' instead of 'count == 0' considering time complexity and readability?
And How about adding the final keyword to classes that do not inherit?

Thanks for reading.

### Checklist

### DIFF
--- a/Sources/DequeModule/Deque._UnsafeHandle.swift
+++ b/Sources/DequeModule/Deque._UnsafeHandle.swift
@@ -188,8 +188,8 @@ extension Deque._UnsafeHandle {
     assert(offsets.lowerBound >= 0 && offsets.upperBound <= count)
     let lower = slot(forOffset: offsets.lowerBound)
     let upper = slot(forOffset: offsets.upperBound)
-    if offsets.count == 0 || lower < upper {
-      return .init(start: ptr(at: lower), count: offsets.count)
+    if offsets.isEmpty || lower < upper {
+    return .init(start: ptr(at: lower), count: offsets.count)
     }
     return .init(first: ptr(at: lower), count: capacity - lower.position,
                  second: ptr(at: .zero), count: upper.position)
--- a/Sources/HashTreeCollections/HashNode/_HashNode+Initializers.swift
+++ b/Sources/HashTreeCollections/HashNode/_HashNode+Initializers.swift
@@ -57,7 +57,7 @@ extension _HashNode {
       childMap: .empty,
       count: 1
     ) { children, items in
-      assert(items.count == 1 && children.count == 0)
+      assert(items.count == 1 && children.isEmpty)
       items.initializeElement(at: 0, to: item)
     }
     r.node._invariantCheck()
@@ -89,7 +89,7 @@ extension _HashNode {
       childMap: .empty,
       count: 2
     ) { children, items -> (_HashSlot, _HashSlot) in
-      assert(items.count == 2 && children.count == 0)
+      assert(items.count == 2 && children.isEmpty)
       let i1 = bucket1 < bucket2 ? 1 : 0
       let i2 = 1 &- i1
       items.initializeElement(at: i1, to: item1)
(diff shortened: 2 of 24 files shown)


Tests/CollectionsTestSupportTests/CombinatoricsChecks.swift
```diff
@@ -14,7 +14,7 @@ import XCTest
 import _CollectionsTestSupport
 #endif
 
-class CombinatoricsTests: CollectionTestCase {
+final class CombinatoricsTests: CollectionTestCase {
```
lorentey, line 17: I'm not interested in microoptimizing XCTestCase classes. Please remove these changes from all test files.

Sources/DequeModule/Deque._UnsafeHandle.swift
```diff
@@ -188,8 +188,8 @@ extension Deque._UnsafeHandle {
     assert(offsets.lowerBound >= 0 && offsets.upperBound <= count)
     let lower = slot(forOffset: offsets.lowerBound)
     let upper = slot(forOffset: offsets.upperBound)
-    if offsets.count == 0 || lower < upper {
```
lorentey, line 191: I don't expect `Range.count == 0` to produce less efficient code than `Range.isEmpty` -- do you have proof that it matters?

Sources/DequeModule/Deque._UnsafeHandle.swift
```diff
@@ -188,8 +188,8 @@ extension Deque._UnsafeHandle {
...
-      return .init(start: ptr(at: lower), count: offsets.count)
```
lorentey, line 192: Please do not break indentation.

Sources/HashTreeCollections/HashNode/_HashNode+Initializers.swift
```diff
@@ -57,7 +57,7 @@ extension _HashNode {
       childMap: .empty,
       count: 1
     ) { children, items in
-      assert(items.count == 1 && children.count == 0)
+      assert(items.count == 1 && children.isEmpty)
```
lorentey, line 60: `children` is of type `UnsafeMutableBufferPointer`. Do you have evidence that this change has a meaningful impact?

lorentey (changes requested):
Thanks for this work! I do not think it is a good idea to start contributing to a project by applying large-scale stylistic changes.

- Please retarget this PR on the `release/1.1` branch. The `future` branch is for experimentation; it isn't intended to ever ship in any production release.

- Marking storage classes final is a good idea. Let's do it! The `_DequeBuffer` change is acceptable, and we should land it. (I expect the compiler infers `final` in this case anyway, but making it explicit is not going to hurt.)

- On the other hand, I am not interested in microoptimizing XCTestCase declarations. Please remove the changes that mark those classes final.

- I do not expect `count == 0` vs `isEmpty` to make any actual difference for the specific types on which we're calling them. Do you have any evidence to the contrary?

- Please make sure not to break indentation.

Tests/CollectionsTestSupportTests/CombinatoricsChecks.swift
quokkaKyu: Thank you for the reply. i will delete the changes in the test files.

Sources/DequeModule/Deque._UnsafeHandle.swift
quokkaKyu: No, I may be mistaken. Since it is a RandomAccessCollection type, I don’t think there will be any difference.

Sources/DequeModule/Deque._UnsafeHandle.swift
quokkaKyu: yes, i won't break the indentation.

Sources/HashTreeCollections/HashNode/_HashNode+Initializers.swift
quokkaKyu: No, I may be mistaken. Since it is a RandomAccessCollection type, I don’t think there will be any difference.

lorentey:
Closing; the `_DequeBuffer` change has landed with #403, and the rest of these changes appear unnecessary. Thanks for this work anyway! If I missed something, please push and reopen as needed.

PR #405, opened 2024-07-15
[BigString] Fix accidentally quadratic `BigString.init`

lorentey:
When ingesting a `String` instance, `BigString` assumes that the input string has a reasonably efficient UTF-8 view.

Unfortunately, that is very much not the case when the input happens to be backed by a bridged NSString object — it appears that in this case, the ingester loop invokes some operation(s) with linear complexity in the size of the entire input, rendering the ingester’s overall complexity quadratic.

The BigString ingester is only expected to operate within a single chunk at the time. It’s unclear precisely which operation triggers the quadratic behavior; ideally we should figure it out and resolve it with a more targeted fix.

In the meantime, a blunt stopgap fix is to force-transcode the input string to UTF-8 at the time the ingester is initialized. This unnecessarily wastes some (temporary) memory on holding the transcoded string, but it avoids the quadratic cliff.

Before:
results-before

After:
results-after

rdar://131176282

### Checklist

### DIFF
--- a/Benchmarks/Sources/Benchmarks/BigStringBenchmarks.swift
+++ b/Benchmarks/Sources/Benchmarks/BigStringBenchmarks.swift
@@ -1,0 +1,108 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Collections open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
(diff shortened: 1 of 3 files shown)


jmschonfeld (approved):

lorentey:
@swift-ci test

PR #414, opened 2024-08-19
[concurrency] conform Deque.Iterator to unchecked-Sendable

glessard:
On a toolchain that contains https://github.com/swiftlang/swift/pull/75782 (79bf5d51ca4b55afdb29102f411898d78cbe0bf2), `Deque` no longer compiles in Swift 6 mode.

`extension Deque._Storage: @unchecked Sendable {}` makes it compile, though it may not be the best solution?

Addresses rdar://134300306

### Checklist


glessard:
@swift-ci please test

Sources/DequeModule/Deque._Storage.swift
```diff
@@ -27,6 +27,8 @@ extension Deque {
   }
 }
 
+extension Deque._Storage: @unchecked Sendable {}
```
lorentey, line 30: Does it make sense to declare `Deque._Storage` sendable, rather than only `Deque` itself?

glessard: There already is a Sendable conformance for `Deque`. I'm not clear why the compiler is tripping over `_Storage`'s sendability status, but setting it separately fixes the error. It should, however, depend on the `Element` type.

glessard:
@swift-ci please test

Sources/DequeModule/Deque._Storage.swift
lorentey: What I mean is that `Deque`'s conformance ought to be `@unchecked`.

lorentey (approved):

Sources/DequeModule/Deque._Storage.swift
glessard: Made `Deque.Iterator`'s conformance unchecked-Sendable; thanks for the nudge.

glessard:
@swift-ci please test

lorentey (approved):

glessard:
@swift-ci please test

PR #447, opened 2025-01-24
fix amd64 support

michael-yuji:
Fix when "${CMAKE_SYSTEM_PROCESSOR}" is lowercase "amd64"


lorentey (changes requested):
This looks good to me, thanks!

I think we'll want to ship this in the near future (in version 1.2); would you please retarget this on top of the `release/1.2` branch, rather than on `main`?

lorentey (approved):

PR #455, opened 2025-03-08
Align Heap._UnsafeHandle min/maxValue tie-breaking with Swift.min/max

### LINKED ISSUE #439: max value decisions are different between Heap.max and Heap.replaceMax
In the Heap module, if the left max value is same as the right max value, the **max** variable will return the right one but the **replaceMax** function will replace the left one

```
struct Fruit: Comparable {
    let name: String
    var price: Int
    
    static func <(lhs: Fruit, rhs: Fruit) -> Bool {
        lhs.price < rhs.price
    }
}

var heap = Heap([
    Fruit(name: "🍎", price: 1),
    Fruit(name: "🍌", price: 10),
    Fruit(name: "🍊", price: 10)
])

if var maxPriceFruit = heap.max {
    maxPriceFruit.price -= 1
    
    heap.replaceMax(with: maxPriceFruit)
}

print(heap.unordered) // [Fruit(name: "🍎", price: 1), Fruit(name: "🍊", price: 9), Fruit(name: "🍊", price: 10)]
```
I think just changing
```
let maxNode = handle.maxValue(.leftMax, .rightMax)
```
to
```
let maxNode = handle.maxValue(.rightMax, .leftMax)
```
(same as **popMax** does) can fix the problem
### Information

- **Package version:** main branch

DakshinD:
Heap._UnsafeHandle has minValue and maxValue. Original functionality was `minValue`, `if a == b, return b`. For `maxValue`, `if a == b, return a`. In both cases, this is the opposite tiebreaker choice from `Swift.min` and `Swift.max`.

`Heap.replaceMax` used `Heap._UnsafeHandle.maxValue` while `Heap.max` used `Swift.max`, and due to different tiebreaker choices, led to inconsistency as seen in #439 .

Therefore I go ahead and update `Heap._UnsafeHandle.maxValue/minValue` to be consistent with `Swift.max/min`.

Fixes #439 

### Checklist

### DIFF
--- a/Sources/HeapModule/Heap+UnsafeHandle.swift
+++ b/Sources/HeapModule/Heap+UnsafeHandle.swift
@@ -81,12 +81,17 @@ extension Heap._UnsafeHandle {
 
   @inlinable @inline(__always)
   internal func minValue(_ a: _HeapNode, _ b: _HeapNode) -> _HeapNode {
-    self[a] <= self[b] ? a : b
+    // The expression used here matches the implementation of the
+    // standard `Swift.min(_:_:)` function. This attempts to
+    // preserve any pre-existing order in case `T` has identity.
+    // `(min(x, y), max(x, y))` should return `(x, y)` in case `x == y`.
+    self[b] < self[a] ? b : a
   }
 
   @inlinable @inline(__always)
   internal func maxValue(_ a: _HeapNode, _ b: _HeapNode) -> _HeapNode {
-    self[a] > self[b] ? a : b
+    //  In case `a` and `b` match, we need to pick `b`. See `minValue(_:_:)`.
+    self[b] >= self[a] ? b : a
   }
 }
 
--- a/Sources/HeapModule/Heap+UnsafeHandle.swift
+++ b/Sources/HeapModule/Heap+UnsafeHandle.swift
@@ -81,12 +81,12 @@ extension Heap._UnsafeHandle {
 
   @inlinable @inline(__always)
   internal func minValue(_ a: _HeapNode, _ b: _HeapNode) -> _HeapNode {
-    self[a] < self[b] ? a : b
+    self[a] <= self[b] ? a : b
   }
 
   @inlinable @inline(__always)
   internal func maxValue(_ a: _HeapNode, _ b: _HeapNode) -> _HeapNode {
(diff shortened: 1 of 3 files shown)


Sources/HeapModule/Heap+UnsafeHandle.swift
```diff
@@ -81,12 +81,12 @@ extension Heap._UnsafeHandle {
 
   @inlinable @inline(__always)
   internal func minValue(_ a: _HeapNode, _ b: _HeapNode) -> _HeapNode {
-    self[a] < self[b] ? a : b
+    self[a] <= self[b] ? a : b
```
lorentey, line 84: I think it would make sense to add a comment explaining that the result emulates the stdlib. The implementation should also match precisely what the stdlib does, to emphasize this even further.

```suggestion
    // The expression used here matches the implementation of the 
    // standard `Swift.min(_:_:)` function. This attempts to 
    // preserve any pre-existing order in case `T` has identity.
    // `(min(x, y), max(x, y))` should return `(x, y)` in case `x == y`.
    self[b] < self[a] ? b : a
```

Sources/HeapModule/Heap+UnsafeHandle.swift
```diff
@@ -81,12 +81,12 @@ extension Heap._UnsafeHandle {
...
   }
 
   @inlinable @inline(__always)
   internal func maxValue(_ a: _HeapNode, _ b: _HeapNode) -> _HeapNode {
-    self[a] < self[b] ? b : a
+    self[a] > self[b] ? a : b
```
lorentey, line 89:
```suggestion
    //  In case `a` and `b` match, we need to pick `b`. See `minValue(_:_:)`.
    self[b] >= self[a] ? b : a
```

lorentey (commented):
Thanks -- this is definitely a good idea.

The goal of this is to guarantee that `Heap` behaves in a specific way for equal-but-distinguishable elements. This new guarantee must come with tests that verify that the implementation behaves as expected; otherwise the bug we want to fix could come back undetected.

DakshinD:
Sounds good, thanks for the revision. Will go ahead and also add those tests.

lorentey:
@swift-ci test

PR #457, opened 2025-04-15
[Future] Introduce Borrow type and use it in Container protocols

Azoy:
With the new toolchain, I've added a `Borrow` type and replaced the unsafe address requirements in favor of safe borrow ones. We also have the ability to turn tests back on and start using `MutableSpan` and thus `MutableContainer`.


lorentey:
🤩

glessard (approved):

PR #458, opened 2025-04-18
Update containers

lorentey:
- Adjust lifetime semantics for `nextSpan`
- Remove `maximumCount` argument from `nextSpan` -- replace it with a default algorithm
- Add `previousSpan` to `BidirectionalContainer`

### DIFF
--- a/Sources/Future/Containers/BidirectionalContainer.swift
+++ b/Sources/Future/Containers/BidirectionalContainer.swift
@@ -20,8 +20,9 @@ public protocol BidirectionalContainer<Element>: Container, ~Copyable, ~Escapabl
 
 @available(SwiftCompatibilitySpan 5.0, *)
 extension BidirectionalContainer where Self: ~Copyable & ~Escapable {
+  @inlinable
   @lifetime(borrow self)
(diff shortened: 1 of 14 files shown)


Sources/Future/Containers/BidirectionalContainer.swift
```diff
@@ -13,4 +13,21 @@
...
   func formIndex(before i: inout Index)
+
+  @lifetime(borrow self)
+  func previousSpan(before index: inout Index) -> Span<Element>
+}
+
+@available(SwiftCompatibilitySpan 5.0, *)
+extension BidirectionalContainer where Self: ~Copyable & ~Escapable {
+  @lifetime(borrow self)
+  func previousSpan(before index: inout Index, maximumCount: Int) -> Span<Element> {
```
glessard, line 24: This still has a `maximumCount` parameter?

glessard (approved):

Sources/Future/Containers/BidirectionalContainer.swift
lorentey: This is the definition of the utility helper that reintroduces it as a default extension. I forgot `@inlinable public` though 🙉

PR #470, opened 2025-05-13
Install swiftmodules with full module triple

etcwilde:
This cleans up how the CMake build of Swift-Collections installs swiftmodules to use the swift module triple expected by the compiler for a given target triple instead of trying to recreate the correct architecture name in CMake. The existing computation is incorrect for modules building for x86_64 FreeBSD machines, causing swift-foundation to fail to build. Letting the compiler specify what it expects the module triple to be for a given target triple will be more robust.

resolves: https://github.com/swiftlang/swift-foundation/issues/1230

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,10 +7,6 @@ Licensed under Apache License v2.0 with Runtime Library Exception
 See https://swift.org/LICENSE.txt for license information
 #]]
 
-if(POLICY CMP0091)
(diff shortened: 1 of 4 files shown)


etcwilde:
@swift-ci please test

compnerd (approved):
This looks great, thanks for fixing this!

lorentey (approved):

PR #486, opened 2025-06-18
[BigString] Harvest some low-hanging performance fruit

lorentey:
This builds on (and temporarily includes the commits of) #485, adds some basic benchmarks, and uses those to make and verify a single set of performance improvements by making use of the known character/scalar/UTF-16 counts to speed up intra-chunk distance calculations.

This makes a roughly 40-50% improvement for distance measurements over `BigString`’s character and Unicode Scalar views. Interestingly, we don’t see an improvement for UTF-16 views — I’m guessing this is due to String’s breadcrumbing. (And it will go away with the upcoming UTF8Spanification of this type.)

```
$ ./Utils/run-benchmarks.sh results compare before.json after.json --output cmp.html
Tasks with difference scores larger than 1.05:
  1.506   1.506   1.513(#48)   0.9925(#8)   BigString.distance(from:to:) (*)
  1.400   1.400   1.405(#51)   0.9950(#5)   BigString.unicodeScalars.distance(from:to:) (*)
2 images written to cmp.html
```

01 BigString distance(from:to:)
02 BigString unicodeScalars distance(from:to:)

(The benchmarks I added only show an improvement for multi-chunk big string instances, but this is merely an artifact of the specific algorithm that is being measured — for single-chunk strings, the old measurement method was already optimal on that payload, so the new heuristic never triggers.)

rdar://153701624

### Checklist

### DIFF
--- a/Sources/RopeModule/BigString/Basics/BigString+Metrics.swift
+++ b/Sources/RopeModule/BigString/Basics/BigString+Metrics.swift
@@ -110,11 +110,24 @@ extension BigString {
       in chunk: BigString._Chunk
     ) -> Int {
       assert(start <= end)
-      return chunk.string.unicodeScalars[start ..< end].count
+
+      // Make use of the chunk's known scalar count to reduce the amount of data we need to look at.
+      if end._utf8Offset - start._utf8Offset < chunk.utf8Count / 2 {
+        return chunk.string.unicodeScalars[start ..< end].count
+      }
+      var result: Int
+      result = chunk.unicodeScalarCount
+      result -= chunk.string.unicodeScalars[..<start].count
+      result -= chunk.string.unicodeScalars[end...].count
+      return result
     }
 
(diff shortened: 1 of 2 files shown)


lorentey:
For reference, here is where we stand versus the standard `String` type:

09 Character-based distance
10 UnicodeScalar-based distance
12 UTF-16 distance
11 UTF-8 distance

It's pretty good, but not quite perfect! `BigString`'s curves very nicely flattens out after a chunk's worth of data.

The beneficial effect of `String`'s breadcrumbs shows up very clear in the UTF-16 baseline.  I don't know why `String`'s UTF-8 view would have a slower `distance` operation than `BigString` (optimizer interference from the bridging paths?), but it's nice we have work to do there too.

lorentey:
(Full results are attached below -- it's a single HTML file)

[results.tar.gz](https://github.com/user-attachments/files/20788213/results.tar.gz)

PR #488, opened 2025-06-25
[BigString] Refactor _Chunk to be its own managed buffer of UTF8

Azoy:
`BigString._Chunk` previously was a struct wrapping a string instance as its backing store. This made the implementation pretty simple because it can defer to string for most string operations.

As part of an effort to show diffs between one big string and another, keeping the chunk representation as is would've meant dipping into its underlying string instance and performing an actual string comparison to get a more finer grained diff. We can utilize many properties of the rope data structure itself to get a very quick answer for a maybe not so precise answer, but chunks not having a stable identity meant we could only operate on the leaves themselves which can contain up to 10 chunks alone (totally at most 2550 utf8 code units which is not precise for a diff).

This patch ditches the string instance for the storage representation and instead makes `_Chunk` a CoW managed buffer of UTF8 bytes. This allows us to have a stable identity that we can compare across big strings to have at worst case a 255 utf8 code unit diff without looking into the strings at all.

### Checklist


Azoy:
Here's a benchmark comparing the performance of the existing `BigString` on `release/1.2` and comparing it to this new implementation: [chart.html.zip](https://github.com/user-attachments/files/20910453/chart.html.zip)

lorentey (approved):

PR #491, opened 2025-06-29
Fix OrderedSet example usage

azarovalex:
Code presented as an example usage of `OrderedSet` doesn't compile since mutating `sort` function is called on immutable `let`.

### Checklist

### DIFF
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet.swift
@@ -39,8 +39,8 @@ import InternalCollectionsUtilities
 /// *in the same order*. This matches the concept of equality of an `Array`, and
 /// it is different from the unordered `Set`.
 ///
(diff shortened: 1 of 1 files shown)


xwu:
On the other hand, a `var` binding that's never mutated triggers a warning, and I think it's a fair argument that we shouldn't give examples that lead to warnings :) It doesn't seem like someone who's at the point of their developer journey that they're looking for third-party packages would realistically have fundamental misunderstandings about `let` vs. `var`.

azarovalex:
> On the other hand, a `var` binding that's never mutated triggers a warning, and I think it's a fair argument that we shouldn't give examples that lead to warnings :) It doesn't seem like someone who's at the point of their developer journey that they're looking for third-party packages would realistically have fundamental misunderstandings about `let` vs. `var`.

@xwu agree with your point, fixed this and a few other places where variables are never mutated but defined with `var`.

lorentey (approved):
Thanks for auditing these!

PR #494, opened 2025-07-17
Add support for embedded Swift mode

parkera:
Remove calls to `debugPrint` (which does not exist), and ifdef out support for `Codable` and `Reflection`.

Note that testing is not available in embedded yet, as far as I know.

### DIFF
--- a/Sources/InternalCollectionsUtilities/Descriptions.swift.gyb
+++ b/Sources/InternalCollectionsUtilities/Descriptions.swift.gyb
@@ -33,6 +33,7 @@ ${modifier} func _addressString<T: AnyObject>(for object: Unmanaged<T>) -> Strin
   _addressString(for: object.toOpaque())
 }
 
+#if !$Embedded
 @inlinable
 ${modifier} func _arrayDescription<C: Collection>(
   for elements: C
@@ -71,5 +72,6 @@ ${modifier} func _dictionaryDescription<Key, Value, C: Collection>(
   result += "]"
   return result
 }
+#endif
 % end
 ${visibility_boilerplate("end")}
--- a/Sources/BitCollections/BitArray/BitArray+Codable.swift
+++ b/Sources/BitCollections/BitArray/BitArray+Codable.swift
@@ -11,6 +11,7 @@
 
 import InternalCollectionsUtilities
 
+#if !$Embedded
 extension BitArray: Codable {
   /// Encodes this bit array into the given encoder.
   ///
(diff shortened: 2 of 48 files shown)


Sources/InternalCollectionsUtilities/Descriptions.swift.gyb
```diff
@@ -45,7 +45,9 @@ ${modifier} func _arrayDescription<C: Collection>(
     } else {
       result += ", "
     }
+#if !$Embedded
     debugPrint(item, terminator: "", to: &result)
+#endif
```
lorentey, line 50: Hm, this replaces `[1, 2, 3]` with `[, , ]` in `description` properties, and that's not very nice.

In this case, the stdlib is currently [substituting a generic message for each item](https://github.com/swiftlang/swift/blob/main/stdlib/public/core/ArrayShared.swift#L124-L128), like this:
```suggestion
#else
    result += "(unprintable)"
#endif
```

This results in `[(unprintable), (unprintable), (unprintable)]`. To be fair, that isn't that useful, either. I think it would be preferable to just replace the entire body with a short summary like `"[\(count) values]"`. We can also choose to entirely omit the `CustomStringConvertible` conformances under `$Embedded`.

(Ditto for `_dictionaryDescription` and all other collection printing code below.)

lorentey (approved):
Looks good, with the nit noted below -- I think we should just `#if` out all `Custom[Debug]StringConvertible` conformances, like we do mirrors and serialization

Sources/InternalCollectionsUtilities/Descriptions.swift.gyb
parkera: Sounds good.

parkera:
> @parkera non-blocking question… do we have build jobs for embedded mode?

I'm not sure if we have integrated embedded support into the GitHub actions  - Swift CI is certainly building toolchains though. cc @shahmishal

Sources/InternalCollectionsUtilities/Descriptions.swift.gyb
parkera: This turns out to be a good idea because the `LosslessStringConvertible` conformance also relied on description.

rauhul:
@parkera thanks for taking such a quick move on this, I'll try to figure out a common CI solution for the swiftlang repos next week, so we can merge this with confidence of it not breaking in the future.

lorentey (approved):

PR #505, opened 2025-08-07
build: Install libraries in an `arch` sub-folder

Steelskin:
This is the proper installation scheme for Swift libraries on Windows and prevents having to manually copy them in `build.ps1`.

### Checklist


Steelskin:
This is meant to be used for the Swift Windows toolchain, which uses version 1.1.6. Please, let me know if this is not the right way to proceed.

lorentey:
@Steelskin ping -- do you really mean this to land only on main?

Steelskin:
> @Steelskin ping -- do you really mean this to land only on main?

My apologies for the late answer. I would very much like to land this on the appropriate release branch. If I understand correctly, this needs to land on main and then be cherry-picked to 1.2 and 1.1?

However, an issue has come up in another similar PR in that other systems do not expect an arch sub-folder. I will modify this PR to add an option to install in an arch sub-folder.

Steelskin:
@lorentey Sorry, this took me a while to get back to. I updated the PR to add the option to install in an arch sub-directory after double-checking that the Linux toolchain installation does not use that layout.

Am I correct in assuming this will need cherry-picking to the 1.2 and 1.1 branches so we can have the changes on Windows?

lorentey:
@Steelskin Yes, this will need to be manually backported to release/1.2 and (possibly) release/1.1 if you need it to apply to any of the earlier release series. I'll also need to schedule new releases for the old branches accordingly.

I'm planning to include this PR in the impending 1.3.0 release, but that will only help adopters who check out specifically that future tag.

PR #535, opened 2025-10-06
[1.1] Update GitHub metadata

lorentey:
- Update pull request and issue templates
- Add workflow configuration

Initially, this simply copies the workflow from `main`. This is not going to work, but I'm curious to see which jobs will actually fail.


lorentey:
> CODE_OF_CONDUCT.md:13:158:* Using welcoming and inclusive language (e.g., prefer non-gendered words like “folks” to “guys”, non-ableist words like “soundness check” to “sanity check”, etc.)

Yeah, no.

PR #616, opened 2026-04-03
[OrderedCollections] Add OrderedDictionary.replaceElement(at:withKey:…

inju2403:
## Summary

Add `replaceElement(at:withKey:value:)` to `OrderedDictionary` and `OrderedDictionary.Elements`.

This method replaces the key-value pair at a given index with a new pair in expected amortized O(1) time.

```swift
var dict: OrderedDictionary = ["a": 1, "b": 2, "c": 3]
let old = dict.replaceElement(at: 1, withKey: "d", value: 4)
// old == (key: "b", value: 2)
// dict is now ["a": 1, "d": 4, "c": 3]
```

## Motivation

`OrderedDictionary` provides O(1) primitives for updating a value by key (`updateValue(_:forKey:)`), replacing a value at an index (`values[index]`), and swapping elements (`swapAt`). However, there is no single operation to replace a key-value pair at a given index when the new key has a different hash than the old one.

| Existing approach | Issue |
|---|---|
| `dict[newKey] = value` | Appends to the end; does not replace at the target index |
| `remove(at:)` + reinsertion | Correct, but O(`count`) due to element shifting |
| `updateValue(_:forKey:)` | Updates the value for an existing key; cannot change the key itself |

Callers can compose `append` → `swapAt` → `removeLast` to work around this, but the pattern is non-obvious and requires careful duplicate-key validation and bounds checking that is easy to omit. `replaceElement(at:withKey:value:)` fills this gap as a first-class operation, consistent with how `swapAt` already exposes an index-based mutation primitive.

> **Note:** If the caller only needs to update the value for the same key at a known index, `values[index] = newValue` provides a direct O(1) alternative.

## Detailed design

**Signature:**

```swift
@discardableResult
public mutating func replaceElement(
  at index: Int,
  withKey key: Key,
  value: Value
) -> Element
```

- Added to both `OrderedDictionary` and `OrderedDictionary.Elements`, matching the existing pattern for index-based mutations.
- Returns the replaced key-value pair (`@discardableResult`).
- Traps if `index` is out of bounds or `key` already exists.
- Complexity: Expected amortized O(1), if `Key` implements high-quality hashing.

**Implementation:** Appends the new pair, swaps it into the target position, and removes the old element from the end — each step is O(1).

## Testing

Tests added in both `OrderedDictionary Tests.swift` and `OrderedDictionary+Elements Tests.swift`, covering:

- Sizes 1 through 19 with every valid index position
- Both unique and shared (copy-on-write) storage via `withHiddenCopies`
- Lifetime tracking to verify no memory leaks
- Return value correctness (old key and value)
- Forward lookup (`dict[newKey] == newValue`)
- Reverse lookup (`dict[oldKey] == nil`)
- Internal invariants via `_checkInvariants()`

### Checklist

### DIFF
--- a/Sources/OrderedCollections/OrderedDictionary/OrderedDictionary+Elements.swift
+++ b/Sources/OrderedCollections/OrderedDictionary/OrderedDictionary+Elements.swift
@@ -396,7 +396,22 @@ extension OrderedDictionary.Elements {
   /// Replaces the key-value pair at the specified index with a new pair,
   /// returning the original element.
   ///
-  /// The new key must not already exist in the dictionary.
+  /// The new key must either not already exist in the dictionary, or be
+  /// equal to the key currently stored at `index`. The latter form is
+  /// useful when equal keys can be distinguished by identity or some other
+  /// means — for example, replacing a decomposed Unicode string with its
+  /// precomposed equivalent:
+  ///
+  ///     var dict: OrderedDictionary = [
+  ///         "a": 1, "e\u{301}": 2, "c": 3
+  ///     ]
+  ///     dict.elements.replaceElement(at: 1, withKey: "é", value: 2)
+  ///     // dict is now ["a": 1, "é": 2, "c": 3]
+  ///
+  /// In the general case, the new pair is appended, swapped into position,
+  /// and the old element is removed from the end — each step is O(1).
+  /// When the new key compares equal to the one being replaced, the pair
+  /// is updated in place and the hash table is left untouched.
   ///
   ///     var dict: OrderedDictionary = [
   ///         "a": 1, "b": 2, "c": 3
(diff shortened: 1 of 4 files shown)


lorentey (commented):
Looks good to me! 

One suggestion: I think we should allow the new key to compare equal to the one being replaced. It's currently rather difficult to (efficiently) replace key instances in OrderedDictionary with equal-but-not-identical substitutes, and this would provide a workable way to do that.

```swift
var dict: OrderedDictionary = [
  "a": 1, "e\u{301}": 2, "c": 3
]
dict.replaceElement(at: 2, withKey: "é", value: 2)
```

My expectation is that this would replace the key `"e\u{301}"` with the key `"é"`, despite them being the "same" as far as hashing is concerned.

inju2403:
@lorentey 
Thanks for the suggestion — the Unicode normalization example is a great motivator for relaxing the precondition.

I've allowed the new key to compare equal to the one already at index; in that case the pair is replaced in place via the existing OrderedSet.update(_:at:), leaving the hash table untouched and preserving the expected amortized O(1) complexity. Genuine duplicates at a different position still trap. Added test_replaceElement_sameKey on both OrderedDictionary and its Elements view to cover the new path.

Thanks again — happy to tweak anything further!

lorentey (approved):
Thanks!

yoonseopshin:
👍🏼 👍🏼 👍🏼

PR #644, opened 2026-05-06
Use the defines from traits directly

FranzBusch:
Package traits already pass defines with their trait name when building. Instead of adding our own conditional defines based on the traits we can use the trait provided defines directly. I observed that various tooling behaves better with the trait provided defines.


lorentey (approved):
Makes sense! 👍

PR #631, opened 2026-04-23
Fix _trim(first:) returning wrong buffer region

FranzBusch:
After reassigning self on line 160, baseAddress on line 161 resolves to the new self's baseAddress (oldStart + cut) instead of the original start. This causes the returned buffer to point past the actual elements into uninitialized memory. Use the captured oldStart instead.

PR #659, opened 2026-05-25
[InternalCollectionsUtilities] Fix _trim returning the wrong buffer region

adityasingh2400:
While reading the buffer-pointer helpers in `InternalCollectionsUtilities`, I noticed that three of the four `_trim` variants compute the trimmed result from `self` after they have already reassigned `self` to the remaining slice. That makes them read the updated `baseAddress`/`count`, so the region they return points into the wrong place in the buffer.

`UnsafeMutableBufferPointer._trim(first:)` does this correctly today: it captures `oldStart` before reassigning `self` and builds the prefix from `oldStart`. (This is the same shape as the earlier fix in #631.) The three siblings were not adjusted to match:

- `UnsafeBufferPointer._trim(first:)` reassigns `self` to `oldStart + cut`, then returns `Self(start: baseAddress, count: cut)`. Since `baseAddress` now points at the remainder, the returned prefix starts `cut` elements too far forward.
- `UnsafeBufferPointer._trim(last:)` and `UnsafeMutableBufferPointer._trim(last:)` set `self` to a buffer of length `count - cut`, then return `Self(start: baseAddress + (count &- cut), count: cut)`. Because `count` has already been reduced by the assignment, the start is computed from the shrunk count and lands `cut` elements too early.

Concretely, for a five-element buffer holding `[10, 20, 30, 40, 50]`:

- `_trim(first: 2)` returned the storage for `[30, 40]` instead of `[10, 20]`.
- `_trim(last: 2)` returned `[20, 30]` instead of `[40, 50]`.

Trimming more than the element count (for example `_trim(first: .max)`) computed a start past the end of the buffer, so the returned span addressed out-of-bounds memory.

The fix captures the original start (and, for the suffix case, the post-trim count) into locals before mutating `self`, so the returned region is derived from the original buffer rather than the already-updated one. This mirrors the existing correct `_trim(first:)` on the mutable buffer.

The mutable `_trim(first:)` overload is reachable today through the deque and array `SubrangeConsumer.drainNext` paths and was already correct; the other three overloads currently have no in-tree callers, but they are part of the same helper family and are wrong as written, so this keeps the set consistent and safe for the next caller that picks them up.

Verification: added direct tests for `_trim(first:)` and `_trim(last:)` on both `UnsafeBufferPointer` and `UnsafeMutableBufferPointer` (plus the over-count and zero-length cases). They fail against the current sources (the immutable `_trim(first:)`, and both `_trim(last:)` overloads, return the wrong elements) and pass with this change. Full `swift test` is green: 802 XCTest cases and 13 swift-testing cases, 0 failures.

### DIFF
--- a/Sources/InternalCollectionsUtilities/UnsafeBufferPointer+Extras.swift
+++ b/Sources/InternalCollectionsUtilities/UnsafeBufferPointer+Extras.swift
@@ -146,7 +146,7 @@ extension UnsafeBufferPointer where Element: ~Copyable {
     guard cut > 0 else { return .init(start: nil, count: 0) }
     let oldStart = baseAddress.unsafelyUnwrapped
     self = Self(start: oldStart + cut, count: count - cut)
-    return Self(start: baseAddress, count: cut)
+    return Self(start: oldStart, count: cut)
   }
 
   @_alwaysEmitIntoClient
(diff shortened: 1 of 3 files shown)


lorentey (approved):
Thank you! 👍

PR #660, opened 2026-05-26
[OrderedCollections] Add move operations

dnadoba:
### Description

Adds new method for relocating elements within an `OrderedSet` and `OrderedDictionary`:
```swift
extension OrderedSet {
  mutating func moveSubrange(_ range: some RangeExpression<Index>, to destination: Index)
  mutating func move(members: some Sequence<Element>, to destination: Index)
  mutating func move(indices: some Sequence<Index>, to destination: Index)
}
extension OrderedDictionary {
  mutating func moveSubrange(_ range: some RangeExpression<Index>, to destination: Index)
  mutating func move(keys: some Sequence<Key>, to destination: Index)
  mutating func move(indices: some Sequence<Index>, to destination: Index)
}
```

This fills a gap that comes up regularly in drag-and-drop UIs and other reordering workflows: today users have to fall back to `remove(at:)` + `insert(_:at:)`, which is O(`count` * *k*) and re-hashes elements unnecessarily.

### Performance

12 new benchmarks in `OrderedSetBenchmarks` cover the public API across the four dispatch paths. The charts below were produced by disabling the fast paths selectively.

Fast path 1: contiguous-range detection in `move(_:Sequence)`. When the input forms a contiguous range, dispatching to the range-rotation path avoids the compact-and-place allocation.

fp1-contiguous-detection

Fast path 2: contiguous-reordered detection. When the sorted sources are contiguous but the input is in a different order, rotation followed by in-place permutation beats compact-and-place.

fp2-contiguous-reordered-detection

Fast path 3: targeted hash updates in `move(_:Range)`. For small range moves over short distances, touching just the affected buckets beats scanning the whole hash table. 

fp3-targeted-hash-range

Fast path 4: targeted hash updates in scattered moves. When the affected region is small relative to table capacity, per-element bucket updates beat a full hash-table rebuild. 

fp4-targeted-hash-scattered

All benchmarks:
all-move-benchmarks

### Source Impact

Additive.

### Checklist

### DIFF
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
@@ -74,32 +74,31 @@ extension OrderedSet {
   }
 
   /// Moves the given elements to the given index, keeping them in the
-  /// order they appear in `elements`.
+  /// order they appear in `members`.
   ///
-  /// Elements of `elements` that are not members of the set are ignored;
-  /// only the members are relocated. `elements` must not contain duplicate
-  /// members.
+  /// Elements that are not members of the set are ignored; only the members
+  /// are relocated. `members` must not contain duplicate elements.
   ///
   ///     var set: OrderedSet = [0, 1, 2, 3, 4, 5, 6]
-  ///     set.move(contentsOf: [4, 1], to: 2)
+  ///     set.move(members: [4, 1], to: 2)
   ///     // set is now [0, 2, 4, 1, 3, 5, 6]
   ///
   /// - Parameters:
-  ///    - elements: The elements to move. Values that are not members of the
+  ///    - members: The elements to move. Values that are not members of the
   ///       set are ignored.
   ///    - destination: The index at which the moved elements should start in
   ///       the resulting set. Must be in the range `0 ... count - k`, where
-  ///       `k` is the number of `elements` that are members of the set.
+  ///       `k` is the number of `members` that are present in the set.
   ///
   /// - Complexity: O(`count`) in the worst case. When the elements form a
   ///    contiguous range or are moved a short distance, the operation is
   ///    proportional to the distance moved.
   @inlinable
   public mutating func move(
-    contentsOf elements: some Sequence<Element>,
+    members elements: some Sequence<Element>,
     to destination: Index
   ) {
-    _move(contentsOf: elements, to: destination)
+    _move(members: elements, to: destination)
   }
 
   /// Moves the elements at the given indices to the given index, keeping them
@@ -108,7 +107,7 @@ extension OrderedSet {
   /// `indices` must contain distinct, valid indices of the set.
   ///
   ///     var set: OrderedSet = [0, 1, 2, 3, 4]
-  ///     set.move(fromIndices: [4, 1], to: 0)
+  ///     set.move(indices: [4, 1], to: 0)
   ///     // set is now [4, 1, 0, 2, 3]
   ///
   /// - Parameters:
@@ -122,16 +121,60 @@ extension OrderedSet {
   ///    proportional to the distance moved.
   @inlinable
   public mutating func move(
-    fromIndices indices: [Index],
+    indices: some Sequence<Index>,
     to destination: Index
   ) {
-    indices.withUnsafeBufferPointer { indices in
-      _move(fromIndices: indices, to: destination)
-    }
+    _move(indices: indices, to: destination)
   }
 }
 
 extension OrderedSet {
+  @inlinable
+  internal mutating func _move(
+    indices: some Sequence<Index>,
+    to destination: Index,
+    applyingTo body: (
+      UnsafeBufferPointer<Int>, UnsafeBufferPointer<Int>, Bool
+    ) -> Void = { _, _, _ in }
+  ) {
+    // Fast path: operate directly on the sequence's storage when available.
+    let handled: Void? = indices.withContiguousStorageIfAvailable { buffer in
+      _move(fromIndices: buffer, to: destination, applyingTo: body)
+    }
+    if handled != nil { return }
+
+    // Fallback: collect the indices into an array, checking bounds and
+    // detecting order as we go.
+    var isContiguousRange = true
+    var isSorted = true
+    var prev = -1
+    var sourceOffsets: [Int] = []
+    sourceOffsets.reserveCapacity(indices.underestimatedCount)
+    for index in indices {
+      precondition(
+        index >= 0 && index < count,
+        "Index \(index) at \(sourceOffsets.count) out of bounds 0 ..< \(count)")
+      if prev >= 0 {
+        if index <= prev {
+          isSorted = false
+          isContiguousRange = false
+        } else if index != prev + 1 {
+          isContiguousRange = false
+        }
+      }
+      prev = index
+      sourceOffsets.append(index)
+    }
+    sourceOffsets.withUnsafeBufferPointer { sourceOffsets in
+      _move(
+        sourceOffsets: sourceOffsets,
+        isContiguousRange: isContiguousRange,
+        isSorted: isSorted,
+        to: destination,
+        applyingTo: body)
+    }
+  }
+
   @inlinable
   internal mutating func _move(
     fromIndices sourceOffsets: UnsafeBufferPointer<Int>,
@@ -167,7 +210,7 @@ extension OrderedSet {
 
   @inlinable
   internal mutating func _move(
-    contentsOf elements: some Sequence<Element>,
+    members elements: some Sequence<Element>,
     to destination: Index,
     applyingTo body: (
       UnsafeBufferPointer<Int>, UnsafeBufferPointer<Int>, Bool
@@ -252,14 +295,11 @@ extension OrderedSet {
       destination >= 0 && destination <= count - c,
       "Destination index \(destination) out of range 0 ... \(count - c)")
 
-    var isNoOp = true
-    for i in 0 ..< c {
-      if sourceOffsets[i] != destination + i {
-        isNoOp = false
-        break
-      }
-    }
-    guard !isNoOp else { return }
+    // A no-op is exactly an in-order contiguous run already sitting at
+    // `destination`. `isContiguousRange` (computed while scanning the input)
+    // already guarantees `sourceOffsets[i] == sourceOffsets[0] + i`, so it
+    // only remains to check that the run starts at `destination`.
+    if isContiguousRange && sourceOffsets[0] == destination { return }
 
     if isSorted {
       // Already ascending: the sorted view is the input view, no copy needed.
--- a/Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
+++ b/Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
@@ -1,0 +1,526 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Collections open source project
+//
+// Copyright (c) 2021 - 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+// SPDX-License-Identifier: Apache-2.0 WITH Swift-exception
+//
+//===----------------------------------------------------------------------===//
+
+#if !COLLECTIONS_SINGLE_MODULE
+import InternalCollectionsUtilities
+#endif
+
+// Reordering implementation.
+//
+// Both `move` overloads relocate elements to a new offset, preserving the
+// order they appear in the input. The work splits into two concerns:
+//
+//   1. Rearrange `_elements` so the moved values are contiguous at the
+//      destination, with the surrounding elements compacted around them.
+//   2. Update `_table` so each element's hash bucket points at its new
+//      offset.
+//
+// The implementation dispatches to one of three buffer strategies depending
+// on the input shape:
+//
+//   * Contiguous range source -> `_moveRange`: a three-reverse rotation of
+//     `_elements`. O(*d* + *k*) element moves, where *d* is the distance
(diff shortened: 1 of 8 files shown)


Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
```diff
@@ -0,0 +1,526 @@
...
+  ///
+  /// - Complexity: O(`count`) in the worst case. When the elements form a
+  ///    contiguous range or are moved a short distance, the operation is
+  ///    proportional to the distance moved.
+  @inlinable
+  public mutating func move(
+    _ elements: some Sequence<Element>,
+    toOffset destination: Int
+  ) {
+    let toMove = ContiguousArray(elements)
```
lorentey: Why not keep `elements` as is? It looks like we're only using `toMove` to iterate over them once, after all; we don't need to copy them into an array up front to do that.

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
```diff
@@ -0,0 +1,526 @@
...
+  ///       `k` is the number of elements being moved.
+  ///
+  /// - Complexity: O(`count`) in the worst case. When the elements form a
+  ///    contiguous range or are moved a short distance, the operation is
+  ///    proportional to the distance moved.
+  @inlinable
+  public mutating func move(
+    _ elements: some Sequence<Element>,
+    toOffset destination: Int
+  ) {
```
lorentey, line 507: There is a naming issue here -- if `Element == Int`, then the first addition (taking `Range<Int>`) becomes a syntactic overload of this generic API, but the two implementations assign very different semantics to their first arguments.

It would be reasonable to expect that the three statements below have consistent behavior:

```
var items: OrderedSet<Int> = ...
items.move(2 ..< 5, toOffset: 8)
items.move(2 ... 4, toOffset: 8)
items.move([2, 3, 4], toOffset: 8)
```

However, the first invocation ends up in the `Range<Int>` "overload", moving _items at offsets_ 2, 3, and 4, as opposed to the other two invocations that move the _items with values_ 2, 3, and 4. The operations are different, and their names should not be confusable.

Additionally, the "toOffset" label on `destination` differs from existing names on this type: e.g., insertion is named `insert(_:at:)`, not `insert(_:atOffset:)`. ("Offset" is not used much in this type at all -- we prefer "index".)

Given the `insert(_:at:)` precedent, the single-element move operation would naturally be called `move(_:to:)`. This suggests that `move(contentsOf:to:)` would be a reasonable name for this operation:

```suggestion
  public mutating func move(
    contentsOf elements: some Sequence<Element>,
    to destination: Index
  ) {
```

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
```diff
@@ -0,0 +1,526 @@
...
+          let prev = sourceOffsets[i - 1]
+          if index <= prev {
+            isSorted = false
+            isContiguousRange = false
+          } else if index != prev + 1 {
+            isContiguousRange = false
+          }
+        }
+        sourceOffsets[i] = index
+      }
```
lorentey: That the implementation immediately maps `elements` to a range set tells me that there is a more primitive operation we should also expose in addition to this one:

```swift
  public mutating func moveSubranges(
    _ subranges: RangeSet<Index>,
    to destination: Index
  )
```

This implementation should then be refactored to construct a `RangeSet`,  rather than just a flat buffer of indices. (I'm hoping here that the stdlib's `RangeSet` provides reasonably efficient construction; that may not in fact be the case.)

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
```diff
@@ -0,0 +1,526 @@
...
+  ///       the resulting set. Must be in the range `0 ... count - range.count`.
+  ///
+  /// - Complexity: O(*d* + *k*) where *d* is the distance between the source
+  ///    and destination, and *k* is the number of elements moved. Falls back
+  ///    to O(`count`) for large moves.
+  @inlinable
+  public mutating func move(
+    _ range: Range<Int>,
+    toOffset destination: Int
+  ) {
```
lorentey: To make this a better fit with our existing operations, I recommend renaming this to:

```suggestion
  public mutating func moveSubrange(
    _ range: Range<Index>,
    to destination: Index
  ) {
```

- Adding a prominent reference to "subrange" matches `replaceSubrange` and `removeSubrange`, strongly hinting that the first argument takes an index range.
- Relabeling the second argument to simply `to` avoids the use of the term "offset". This type prefers to call these "indices", with only prepositions as labels. (`insert(_:at:)`, `remove(at:)`, etc.), following the scheme established by `RangeReplaceableCollection`.
- Instead of `Int`, I find it's actually clearer to use the typealias `Index` here, to help explain the precise role of these integers. (This is a break from the other operations on this type -- I used to prefer not using typealiases, to avoid forcing users to look up what type they resolve to.)

lorentey: We should also have a generic overload of this that takes a `RangeExpression`:

```swift
public mutating func moveSubrange(
  _ range: some RangeExpression<Index>,
  to destination: Index
)
```

lorentey (commented):
Nicely done! I have some naming recommendations and a refactoring suggestion

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
dnadoba: Partially a habit of writing code that doesn't get specialized 😅

I currently need to access `count` upfront to allocate enough memory in `withUnsafeTemporaryAllocation`.
If we upgrade from requirement from `Sequence` to `Collection` that should work but sometimes `count` is still linear and does a full iteration so it is a tradeoff.

How do you feel about using `Collection` and accessing `count` on it?

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
dnadoba: That is a really good catch! This didn't come to my mind at all. Thanks for thinking this through in detail, I will take your suggestions.

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
dnadoba: It maps it to indices, not range sets. 
The order of the indices here is important. `RangeSet` would lose the order.
For example:
```swift
var set: OrderedSet = [0, 1, 2, 3, 4, 5, 6]
set.move([4, 1], toOffset: 2)
```
Here we want to get `[0, 2, 4, 1, 3, 5, 6]` and **not** `[0, 2, 1, 4, 3, 5, 6]` but a `RangeSet` is always sorted ascending and would only really be able to do the later. 

I think the more fundamental operations is based on a collection of indices. I'm currently working on bringing the same operation to `OrderedDictionary` where I'm refactoring this to be more reusable as well as I need to do the same move operations to the values Array. The catch here is that while gathering the indices we can essentially for free gather information that allows us to dispatch to fast paths (the `isContiguousRange` and `isSorted` booleans from above). A more fundamental API that accepts a collection of indices would need to do a full iteration to recover that information. In the internal variant I'm currently passing additional arguments that makes it possible to dispatch to fast paths. A public variant would likely skip those fast paths and users would use the `Range[Expression]` variants if they really want to hit the fast paths.

Additionally `RangeSet` is generic but largely not marked `@inlinable` nor `@frozen` and doesn't specialize well if at all. The overhead is significant from past experience and not usable (rdar://146195611).

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
lorentey: Huh. The thing that rubs me the wrong way is that we're actively copying items into a heap-allocated array, only to have a precise count to feed to the temporary allocation. I get that `_copyToContiguousArray` works well in unspecialized contexts, but to me it seems unlikely that copying the items like this would be a good idea.

I'd much rather see us simply map the elements directly into an index array, avoiding having to copy them:

```swift
let indices = elements.map { item in
  guard let index = _find(item).index else {
      preconditionFailure(
        "Attempted to move an element that is not a member of the set")
        }
  ... etc
}
self.move(elementsAtIndices: indices, to: destination)
```

There is no technical reason we can't implement chunked iteration on `Sequence`; but even with elementwise iteration, I expect the cost of this loop to be dominated by all the hashing in `_find`.

dnadoba: That makes sense. So you would remove `withUnsafeTemporaryAllocation` in favor of an Array?
Works for me but just want to double check.

dnadoba:
Update public API:

```swift
extension OrderedSet {
  public mutating func moveSubrange(_ range: some RangeExpression<Index>, to destination: Index)
  public mutating func move(contentsOf elements: some Sequence<Element>, to destination: Index)
  public mutating func move(fromIndices indices: [Index], to destination: Index)
}
```

I also recalled an older conversation with my team and update `move(contentsOf:)` to ignore elements/keys that aren't present. That can sometimes happen during a drag & drop operation that races with a concurrent mutation from e.g. a network sync from other devices or other users. 

I have also added another fast path. When a scattered move requires a full hash-table rebuild and the `OrderedSet`'s table storage is *not* uniquely referenced, it builds a fresh table from scratch instead of copying the shared table only to immediately clear and refill it.

While refactoring it was now also trivial to add API to `OrderedDictionary`:

```swift
extension OrderedDictionary {
  public mutating func moveSubrange(_ range: some RangeExpression<Index>, to destination: Index)
  public mutating func move(contentsOf keys: some Sequence<Key>, to destination: Index)
  public mutating func move(fromIndices indices: [Index], to destination: Index)
}
```


`OrderedSet` resolves the move and calls back through an internal closure with the resolved offset views, and the dictionary applies the same rearrangement to its values in lockstep with the keys. It has to be a closure rather than returning the offsets because those offset buffers are stack-allocated via `withUnsafeTemporaryAllocation` and can't escape the call. It also has to be a non-optional closure as optional closures become escaping closures and then the use site can't capture the values inout. 

I wasn't too sure about the `fromIndices ` parameter label. Let me know if you have a better idea and I can rename it.

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
```diff
@@ -0,0 +1,683 @@
...
+  ///    - destination: The index at which the moved elements should start in
+  ///       the resulting set. Must be in the range `0 ... count - k`, where
+  ///       `k` is the number of elements being moved.
+  ///
+  /// - Complexity: O(`count`) in the worst case. When the elements form a
+  ///    contiguous range or are moved a short distance, the operation is
+  ///    proportional to the distance moved.
+  @inlinable
+  public mutating func move(
+    fromIndices indices: [Index],
```
dnadoba: I have also though about using `Span<Index>` here which would be trivial to change. Not sure what the current API guidelines are for this kind of API. Sequence is tricky here as we iterate the array currently twice. Once to check if it is already sorted and if not we sort it.

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
```diff
@@ -0,0 +1,683 @@
...
+    indices.withUnsafeBufferPointer { indices in
+      _move(fromIndices: indices, to: destination)
+    }
+  }
+}
+
+extension OrderedSet {
+  @inlinable
+  internal mutating func _move(
+    fromIndices sourceOffsets: UnsafeBufferPointer<Int>,
```
dnadoba, line 223: Generally I could also replace all (or most) `UnsafeBufferPointer` usage with `Span` internally. 
I shortly experimented and it seemed that at least with Swift 6.4 I didn't run into any limitations and it seemed to work just fine.

Downside is that it will require at least Swift 6.2+, maybe higher depending on if we need any API that came out later.
What's your feeling here? I don't think anything else in this module uses Span so it might be a bit inconstant.

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
dnadoba: I ended up using Sequence and added two paths. 
1. A fast path that uses withContiguousStorageIfAvailable + withUnsafeTemporaryAllocation. 
2. A slow path that uses normal Sequence iteration + an Array as temporary resizable storage.

I shortly also just used map{} but when I also added support for missing elements I would need to use compactMap which doesn't reserve capacity upfront. I'm expecting that most of the time we find all elements and so I'm reserving enough capacity for `elements.underestimatedCount` which should give us decent performance in this case as well.

The fast path is really required for unspecialized code. The iterator from `Array` (which is just the default `IndexingIterator` so a lot of other collections are affected) calls into Arrays read subscript accessor which allocates per access. As a bonus, debug builds benefits from this as well as they are always unspecialized.

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
lorentey: Oh; this absolutely needs to be generic over `Sequence` -- but it makes sense to include this version as a fast path overload or specialization of the primary generic API

dnadoba: Done

Sources/OrderedCollections/OrderedSet/OrderedSet+Move.swift
lorentey: This is fine!

`Span` has SwiftStdlib 5.0 availability and that prevents us from migrating existing code to use it until the minimum deployment target of our oldest supported toolchain grows beyond that. (New API additions can start using it right now, though; they would just need to be marked with `@available(SwiftStdlib 5.0, *)`.

lorentey (approved):

PR #662, opened 2026-05-31
Run heap node tests in release

### LINKED ISSUE #101: [Heap] We should be able to run tests in optimized builds
Currently the `Heap` tests in [./Tests/PriorityQueueTests/HeapTests.swift](https://github.com/apple/swift-collections/blob/main/Tests/PriorityQueueTests/HeapTests.swift) are relying on `@testable import` to access internal interfaces. Unfortantely this doesn't work in release mode, so we can only run heap tests in debug builds.

We ought to be able to run these correctness tests with optimizations enabled. The way to do this is to replace direct use of internal interfaces with an `__unstable` view and/or `@_spi(Testing)` -- see e.g. how `OrderedSet` does this.

sorinc03:
## Summary
- add a Testing SPI helper for heap level calculation checks
- switch HeapNodeTests from debug-only @testable access to @_spi(Testing) imports
- include the helper in the CMake source list

Fixes #101.

## Validation
- swift test --configuration release --filter HeapNodeTests
- swift test --filter HeapNodeTests
- git diff --check

### DIFF
--- a/Sources/HeapModule/Heap+Testing.swift
+++ b/Sources/HeapModule/Heap+Testing.swift
@@ -11,9 +11,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(Testing)
-public enum _HeapTestSupport {
-  public static func isMinLevel(offset: Int) -> Bool {
+extension Heap {
+  package static func _isMinLevel(offset: Int) -> Bool {
     _HeapNode(offset: offset).isMinLevel
   }
 }
--- a/Sources/HeapModule/Heap+Testing.swift
+++ b/Sources/HeapModule/Heap+Testing.swift
@@ -1,0 +1,19 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Collections open source project
+//
+// Copyright (c) 2021 - 2026 Apple Inc. and the Swift project authors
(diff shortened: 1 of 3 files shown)


Sources/HeapModule/Heap+Testing.swift
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH Swift-exception
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(Testing)
+public enum _HeapTestSupport {
+  public static func isMinLevel(offset: Int) -> Bool {
+    _HeapNode(offset: offset).isMinLevel
+  }
+}
```
lorentey: Let's just put this on `Heap`, and let's also migrate `@testable`/`@_spi(Testing)` to the (much simpler) `package` modifier!

```suggestion
extension Heap {
  package static func _isMinLevel(offset: Int) -> Bool {
    _HeapNode(offset: offset).isMinLevel
  }
}
```

lorentey (commented):
👍 

Arguably that test doesn't carry much, but as long as we keep it, we may as well do it properly.

lorentey (approved):
👍

PR #668, opened 2026-06-12
[OrderedCollections] OrderedSet.reverse(): update the hash table in place

inju2403:
## Summary

`OrderedSet.reverse()` now updates its hash table in place instead of rebuilding it from scratch. Behaviour is identical; reversing large sets is roughly 2× faster.

## Motivation

While tracing how `OrderedSet` keeps its hash table consistent across mutations — the `bias` trick in `adjustContents`, the bucket swaps in `swapAt` / `partition` / the move APIs — I noticed `reverse()` is the odd one out: it throws the whole table away and rebuilds it. The code has flagged this since the original 2021 implementation:

```swift
public mutating func reverse() {
  _elements.reverse()
  // FIXME: Update hash table contents in place.
  _regenerateHashTable()
  _checkInvariants()
}
```

Reversing is a deterministic permutation: the element at offset `i` moves to `count - 1 - i`. Hash values are unaffected, so every element stays in its existing bucket — only the *stored offset* in each bucket needs to change.

| Approach | Cost |
|---|---|
| `_regenerateHashTable()` (current) | Allocates a new table and recomputes every element's hash |
| In-place offset flip (this PR) | Walks the buckets once, rewriting `v -> count - 1 - v`; no hashing, no allocation |

## Detailed design

Adds an internal `_UnsafeHashTable.reverse(count:)` that walks the buckets and rewrites each stored offset `v` to `count - 1 - v`, using the same bucket-walk idiom as the existing `adjustContents` / `swapBucketValues` helpers. `OrderedSet.reverse()` invokes it through a new `_reverseHashTable()` (shaped exactly like `_regenerateExistingHashTable()`), replacing the full-rebuild call.

- No public API change; `reverse()`'s behaviour and complexity (O(`count`)) are unchanged.
- The flip stays within the table's representable offset range, since it only permutes the existing set of stored offsets.

**Measurements** — reversing a 1,000,000-element set (release, mean of 50):

| Element | Before (rehash) | After (in-place) | Speedup |
| --- | ---: | ---: | ---: |
| `OrderedSet<Int>` | 72.0 ms | 33.1 ms | ~2.2× |
| `OrderedSet<String>` | 158.3 ms | 57.4 ms | ~2.8× |

The win is larger for expensive-to-hash elements, which pay the most for the avoided rehash. (`reverse()` is an explicit O(n) bulk operation; `reversed()` returns a lazy view that doesn't go through it.)

## Testing

- The existing `test_reverse` already exercises the change across `withOrderedSetLayouts(scales: [0, 5, 6])` (every interesting `scale`/`bias`) and shared/unshared storage; its `_checkInvariants()` confirms every element resolves to its correct offset.
- Added `test_reverse_lookups`, asserting `firstIndex(of:)` / subscript for every element (plus a negative lookup) after `reverse()`. Unlike `_checkInvariants()`, these lookups validate hash-table integrity regardless of `COLLECTIONS_INTERNAL_CHECKS`.
- Added an `OrderedSet<Int> reverse` benchmark.
- All 101 `OrderedSetTests` pass with and without `-Xswiftc -DCOLLECTIONS_INTERNAL_CHECKS`.

### Checklist

### DIFF
--- a/Benchmarks/Libraries/OrderedSet.json
+++ b/Benchmarks/Libraries/OrderedSet.json
@@ -38,6 +38,7 @@
             "OrderedSet<Int> random swaps",
             "OrderedSet<Int> partitioning around middle",
             "OrderedSet<Int> sort",
+            "OrderedSet<Int> reverse",
           ]
         },
         {
--- a/Benchmarks/Sources/Benchmarks/OrderedSetBenchmarks.swift
+++ b/Benchmarks/Sources/Benchmarks/OrderedSetBenchmarks.swift
@@ -131,6 +131,19 @@ extension Benchmark {
       }
     }
 
+    self.add(
+      title: "OrderedSet<Int> reverse",
+      input: [Int].self
+    ) { input in
+      return { timer in
+        var set = OrderedSet(input)
+        timer.measure {
+          set.reverse()
+        }
+        blackHole(set)
+      }
+    }
+
     self.addSimple(
       title: "OrderedSet<Int> append",
       input: [Int].self
--- a/Sources/OrderedCollections/HashTable/_HashTable+UnsafeHandle.swift
+++ b/Sources/OrderedCollections/HashTable/_HashTable+UnsafeHandle.swift
@@ -496,6 +496,20 @@ extension _UnsafeHashTable {
 }
 
 extension _UnsafeHashTable {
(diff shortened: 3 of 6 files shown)


lorentey (approved):
🚀

PR #671, opened 2026-06-22
Fix Collections reexports under MemberImportVisibility

### LINKED ISSUE #663: `import Collections` does not work as expected since 1.4.0 when `MemberImportVisibility` is enabled
The README says that importing `Collections` is enough to use types such as `OrderedSet`.

However, since 1.4.0, when `MemberImportVisibility` is enabled, array literal syntax and method calls on these types do not work. Users need to import the individual defining modules, such as `OrderedCollections`, as well.

I believe this is caused by the export mechanism changing to `typealias` in 1.4.0. This appears to be a source-breaking change under `MemberImportVisibility`.

### Information

- **Package version:** Since 1.4.0
- **Platform version:** macOS Tahoe 26.2
- **Swift version:** swift-driver version: 1.148.6 Apple Swift version 6.3.2 (swiftlang-6.3.2.1.108 clang-2100.1.1.101) Target: arm64-apple-macosx26.0

### Checklist

### Steps to Reproduce

Here is a minimal repro.

Manifest:

```swift
// swift-tools-version: 6.0
import PackageDescription

let package = Package(
    name: "MemberImportVisibilityRepro",
    platforms: [.macOS(.v15)],
    products: [
        .library(name: "Repro", targets: ["Repro"]),
    ],
    dependencies: [
        .package(url: "https://github.com/apple/swift-collections.git", exact: "1.4.0"),
    ],
    targets: [
        .target(
            name: "Repro",
            dependencies: [
                .product(name: "Collections", package: "swift-collections"),
            ],
            swiftSettings: [
                .enableUpcomingFeature("MemberImportVisibility"),
            ]
        ),
    ]
)
```

Source:

```swift
import Collections

public struct Repro {
    private var values: OrderedSet<String> = []
    // ERROR: Initializer 'init(arrayLiteral:)' is not available due to missing import of defining module 'OrderedCollections'

    public init() {}

    public mutating func insert(_ value: String) -> Int {
        values.append(value).index
        // ERROR: Instance method 'append' is not available due to missing import of defining module 'OrderedCollections'
    }
}
```

### Expected behavior

No errors occur.

### Actual behavior

Errors occur.

kiwigitops:
## Summary

- re-export component modules from the aggregate Collections module
- run CollectionsModuleTests with the repo's upcoming feature settings
- add an OrderedSet member/import visibility regression

Fixes #663

## Testing

- `git diff --check`
- Attempted `swift test --filter CollectionsTests/testOrderedSetMemberImportVisibility`; local Swift 6.1.2 cannot parse the current package manifest because it uses newer manifest APIs

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -361,7 +361,7 @@ let targets: [CustomTarget] = [
     kind: .test,
     name: "CollectionsModuleTests",
     dependencies: ["Collections", "_CollectionsTestSupport"],
-    settings: _baseSettings),
+    settings: _testSettings),
 ]
 
 let _products: [Product] = targets.compactMap { t in
--- a/Sources/Collections/BitCollections reexports.swift
+++ b/Sources/Collections/BitCollections reexports.swift
@@ -12,7 +12,7 @@
 //===----------------------------------------------------------------------===//
 
 #if !COLLECTIONS_SINGLE_MODULE
-import BitCollections
+@_exported import BitCollections
 
 public typealias BitSet = BitCollections.BitSet
 public typealias BitArray = BitCollections.BitArray
--- a/Sources/Collections/DequeModule reexports.swift
+++ b/Sources/Collections/DequeModule reexports.swift
@@ -12,7 +12,7 @@
 //===----------------------------------------------------------------------===//
 
 #if !COLLECTIONS_SINGLE_MODULE
-import DequeModule
+@_exported import DequeModule
 
 public typealias Deque<Element> = DequeModule.Deque<Element>
 #endif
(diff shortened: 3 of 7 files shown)


lorentey (changes requested):
I stopped using `@_exported import` in #566 to have the package stop relying on this non-public, unstable attribute, and to avoid dumping the new `RigidDeque` and `UniqueDeque` types on everyone importing `Collections`.

We can discuss reverting that change, but reinstating `@_exported imports` while leaving the typealiases in place is not the way to do that.

PR #674, opened 2026-06-23
Fix superset set algebra doc comments

wjdtjq6:
Fixes documentation comments for set-algebra superset APIs where several overloads described the result as a subset check, or used examples that did not match the overload.

This updates BitSet, BitSet.Counted, TreeSet, and OrderedSet.UnorderedView comments so isSuperset and isStrictSuperset descriptions/examples match the implemented API semantics.

Documentation only; no behavior changes.


lorentey (approved):
Thank you!

PR #679, opened 2026-06-29
Fix @inlinable deinit incompatibility with library evolution (HashTree/Sorted storage)

### LINKED ISSUE #676: HashTreeCollections: @inlinable deinit on _HashNode incompatible with BUILD_LIBRARY_FOR_DISTRIBUTION=YES (library evolution mode)
## Summary

When building a Swift package that depends on `swift-collections` with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` (library evolution / resilient ABI mode) on Xcode 26, the build fails because `_HashNode`'s `deinit` is marked `@inlinable` but the class is not `@_fixed_layout`.

This blocks SDK vendors who distribute binary XCFrameworks — `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` is required to emit `.swiftinterface` files for cross-compiler-version compatibility, but it cannot be used alongside `swift-collections` 1.4.0 or later.

## Error

```
_HashNode+Storage.swift:47:5:
error: deinitializer can only be '@inlinable' if the class is '@_fixed_layout'
```

Relevant source (`Sources/HashTreeCollections/HashNode/_HashNode+Storage.swift`, line 47, present through `main` as of 2026-06-26):

```swift
@inlinable
deinit {
    handle.children.deinitialize()
    handle.reverseItems.deinitialize()
}
```

## Root Cause

- `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` enables library evolution (resilient ABI): class instance layout becomes opaque to external modules unless the class is marked `@_fixed_layout`.
- `@inlinable deinit` requires the caller to inline the deinitializer instructions directly, which in turn requires the class layout to be known at compile time.
- `_HashNode` is a resilient class (no `@_fixed_layout`), so the combination is rejected by the Swift compiler.

This is not a bug in `_HashNode`'s logic — it is a fundamental incompatibility between `@inlinable` on a deinit and resilient layout. The compiler diagnostic enforcing this was tightened in Xcode 26 (Swift 6.2); the same code compiles without error on Xcode 16.4 (Swift 6.1).

## Reproduction

1. Create a Swift package that depends on `swift-collections` 1.4.0 or later.
2. Build with library evolution enabled:

```bash
xcodebuild build \
  -scheme YourScheme \
  -destination "generic/platform=iOS" \
  BUILD_LIBRARY_FOR_DISTRIBUTION=YES \
  CODE_SIGNING_ALLOWED=NO
```

3. Observe the error on `_HashNode+Storage.swift:47`.

## Environment

- Xcode 26.3 / Swift 6.2
- swift-collections 1.4.0, 1.4.1, 1.5.0, 1.5.1, 1.6.0 (all affected; `@inlinable deinit` present in `main` as of 2026-06-26)
- swift-collections 1.3.0: **not affected** (no `@inlinable deinit` on `_HashNode`)
- Xcode 16.4 / Swift 6.1: **does not reproduce** (compiler diagnostic not enforced)

## Impact

SDK vendors who:
- Use `swift-collections` as a transitive dependency
- Must produce binary XCFrameworks with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES`

…are hard-pinned to `swift-collections` 1.3.0 and cannot pick up any newer release.

## Possible Fix

One of:
- Remove `@inlinable` from `_HashNode.deinit` (trades inlinability for binary-distribution compatibility)
- Mark `_HashNode` as `@_fixed_layout` if its layout is truly stable (though this is a strong ABI commitment)
- Guard the `@inlinable deinit` behind a compiler version check if the intent is to keep it for non-library-evolution builds only

itej13:
### Problem

Building a package that depends on swift-collections with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` (library evolution) fails on Xcode 26 / Swift 6.2+:

```
_HashNode+Storage.swift:47:5: error: deinitializer can only be '@inlinable' if the class is '@_fixed_layout'
```

The `Storage` class in HashTreeCollections marks its `deinit` as `@inlinable` but lacks `@_fixed_layout`. Inlining a `deinit` requires a known layout, which library evolution deliberately hides — so the two are mutually exclusive and the build is rejected. This blocks SDK vendors distributing binary XCFrameworks from using 1.4.0+.

### Fix

Add `@_fixed_layout` to the affected internal storage classes, mirroring what is **already done** for the equivalent `ManagedBuffer` storage classes in `DequeModule` (`_DequeBuffer`) and `RopeModule` (`Rope._Storage`). As the comment on `Rope._Storage` states, this does **not** make anything ABI stable — these classes are `@usableFromInline internal` and never part of the public ABI surface. It only allows the modules to build with library evolution enabled.

The same latent issue exists in **SortedCollections** (`_Node.Storage`), which has the identical `@inlinable deinit` without `@_fixed_layout`. A vendor enabling library evolution package-wide hits HashTreeCollections first and SortedCollections immediately after, so both are fixed here. (Happy to split into a separate PR if preferred.)

### Verification

Reproduced the diagnostic with a minimal case at the exact visibility of both classes (`@usableFromInline` class + `@inlinable deinit` + `-enable-library-evolution`) — the error fires before the change and compiles cleanly after. Both modules continue to build normally.

Fixes #676.

### Checklist
- [x] I've read the Contribution Guidelines
- [x] My contributions are licensed under the Swift license.
- [ ] I've added tests covering all new code paths (this is a build-mode constraint with no runtime behavior to test; CI does not currently exercise library-evolution mode).
- [ ] I've added benchmarks (n/a).
- [ ] I've updated the documentation if necessary (n/a).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/HashTreeCollections/HashNode/_HashNode+Storage.swift
+++ b/Sources/HashTreeCollections/HashNode/_HashNode+Storage.swift
@@ -37,6 +37,7 @@ extension _HashNode {
   /// Instances of this class hold (tail-allocated) storage for individual
   /// nodes in a hash tree.
   @usableFromInline
+  @_fixed_layout // Not really! This module isn't ABI stable.
   internal final class Storage: _RawHashStorage {
     @usableFromInline
     internal typealias Element = (key: Key, value: Value)
--- a/Sources/SortedCollections/BTree/_Node.Storage.swift
+++ b/Sources/SortedCollections/BTree/_Node.Storage.swift
@@ -117,7 +117,8 @@ extension _Node {
   /// does not exist, ``_Node.dummyValue`` can be used to obtain a valid value within the type system
   /// that can be passed to APIs.
   @usableFromInline
-  internal class Storage: ManagedBuffer<_Node.Header, Key> {    
+  @_fixed_layout // Not really! This module isn't ABI stable.
+  internal class Storage: ManagedBuffer<_Node.Header, Key> {
     /// Allows **read-only** access to the underlying data behind the node.
     ///
     /// - Parameter body: A closure with a handle which allows interacting with the node


lorentey (approved):
Thanks! 👍

PR #684, opened 2026-07-15
StringCollection

CTMacUser:
Since `String` objects use dynamic memory, any collection of them that also uses dynamic memory will use up several patches of dynamic allocation. I wondered if we can reduce that memory allocation to one buffer. This is funnily similar to how `String` objects could be made up of UTF-8 bundles of data.

UTF-8 lay outs its data such that we can naturally find borders between code points. `String` objects can't have that, so I had to use marker values to indicate where a string's embedded data starts and ends, bloating the total size of the data stored.

Since a `StringCollection` stores elements with varying lengths, it cannot support `RandomAccessCollection` nor `MutableCollection`. Like `String`, `StringCollection` only supports up to `BidirectionalCollection` and `RangeReplaceableCollection`.

`StringCollection` uses a copy-on-write policy on its mutating operations. If a collection the only user of a string data block, the block will be manipulated directly, otherwise a computed clone is made.

`StringCollection` follows many other system operations (`Comparable`, `CustomStringConvertible`, `Decodable`, `Encodeable`, `Equatable`, `ExpressibleByArrayLiteral`, and `Hashable`).

### Checklist


MahdiBM:
I think it's worth linking to the Swift forums thread: https://forums.swift.org/t/a-memory-saving-type-for-collections-of-strings

PR #692, opened 2026-07-22
[RigidArray/UniqueArray] Split up the buffer pointer in RIgidArray

Azoy:
This changes `RigidArray`'s implementation from storing an `UnsafeMutableBufferPointer` to an `UnsafeMutablePointer` and `Int` capacity. Buffer pointer stores an optional pointer which unfortunately doesn't leave us with a bit that `Optional` can use when paired with `UniqueArray?` (this has a size of 25 and a stride of 32 vs. the expected 24 size and 24 stride). We can store a non-optional pointer and put a sentinel well aligned value in there when we want to lazily allocate the array. If `_capacity == 0` then the pointer value in the array is effectively nil.

### Checklist


lorentey (approved):
Thanks! 👍

PR #697, opened 2026-08-02
Fix inverted SortedDictionary Keys/Values/SubSequence equality

### LINKED ISSUE #696: SortedDictionary.Keys.== and .Values.== use == where != was intended, inverting the comparison
`SortedDictionary.Keys` and `SortedDictionary.Values` both implement `==` with an
inverted element comparison, so the operator returns the opposite of the intended
result for any non-empty view.

Reproduced on `main` at `ff27e367`.

### Affected code

Both bodies bail out when a pair of elements is **equal**:

https://github.com/apple/swift-collections/blob/ff27e367/Sources/SortedCollections/SortedDictionary/SortedDictionary%2BKeys.swift#L239-L247

https://github.com/apple/swift-collections/blob/ff27e367/Sources/SortedCollections/SortedDictionary/SortedDictionary%2BValues.swift#L258-L266

```swift
public static func ==(lhs: Self, rhs: Self) -> Bool {
  if lhs.count != rhs.count { return false }
  for (e1, e2) in zip(lhs, rhs) {
    if e1 == e2 {        // `!=` intended
      return false
    }
  }
  return true
}
```

`SortedSet.==`, `SortedSet.SubSequence.==`, `SortedDictionary.==` and
`SortedDictionary.SubSequence.==` use `!=` and behave correctly — only these two
projection views are affected.

### Reproduction

`SortedCollections` is behind the `UnstableSortedCollections` trait, so a
reproducer has to opt in:

```swift
// Package.swift
dependencies: [
  .package(
    url: "https://github.com/apple/swift-collections.git",
    branch: "main",
    traits: ["UnstableSortedCollections"]
  )
]
```

```swift
import SortedCollections

var dict = SortedDictionary<Int, String>()
for index in 1...3 { dict[index] = "v\(index)" }

var other = SortedDictionary<Int, String>()
for index in 4...6 { other[index] = "w\(index)" }

print(dict.keys == dict.keys)      // false — expected true
print(dict.values == dict.values)  // false — expected true
print(dict.keys == other.keys)     // true  — expected false ([1,2,3] vs [4,5,6])

print(dict == dict)                // true  (control — SortedDictionary is correct)
```

### Actual vs expected

| expression | actual | expected |
|---|---|---|
| `dict.keys == dict.keys` | `false` | `true` |
| `dict.values == dict.values` | `false` | `true` |
| `dict.keys == other.keys`, disjoint, same count | `true` | `false` |

So the operator is not merely non-reflexive: two views with **no elements in
common** compare equal, provided they have the same count.

### Knock-on effect in hashed containers

Both views conform to `Hashable`, so the broken `==` propagates:

```swift
let keys = dict.keys
print(Set([keys]).contains(keys))   // false — a value cannot find itself

var lookup: [SortedDictionary<Int, String>.Keys: String] = [:]
lookup[keys] = "hello"
print(lookup[keys] as Any)          // nil — the inserted value is unreachable
lookup[keys] = "again"
print(lookup.count)                 // 2 — a duplicate key was inserted
```

### Suggested fix

One character in each file:

```diff
-      if e1 == e2 {
+      if e1 != e2 {
```

`SortedDictionary.Values: Equatable` is already constrained `where Value: Equatable`,
so `!=` is available there; `Keys`' element is `Key`, which is `Comparable`.

I applied exactly this to a checkout of `ff27e367` and re-ran the reproducer above:
all three expressions, and all three hashed-container cases, produce the expected
results afterwards.

### Test coverage

`Tests/SortedCollectionsTests/` has no equality test for either view, which is
why this survives. `_CollectionsTestSupport`'s `checkEquatable` battery would
catch it immediately — reflexivity fails on the first non-empty instance — so
adding these two views to the existing `check*` coverage would both fix the gap
and guard it.

### Scope

Filing for completeness rather than urgency: `UnstableSortedCollections` is
commented out of the default trait set, and the trait description states these
types are early drafts not intended for production, so no released configuration
is affected. Happy to close if this is already covered by the `SortedSet` /
`SortedDictionary` work tracked in #1.

*Found while surveying `==` implementations across Swift collection libraries.*

Hashim1999164:
## Summary
- Fixes inverted element comparisons in `SortedDictionary.Keys`, `.Values`, and `.SubSequence` equality (`==` was used where `!=` was intended).
- Adds focused Equatable coverage for those views.
- Resolves #696.

## Test plan
- [x] `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --traits UnstableSortedCollections --filter 'SortedDictionaryTests.test_(keys|values|subSequence)_Equatable'`
- [ ] CI on this PR

### DIFF
--- a/Sources/SortedCollections/SortedDictionary/SortedDictionary+SubSequence.swift
+++ b/Sources/SortedCollections/SortedDictionary/SortedDictionary+SubSequence.swift
@@ -300,7 +300,7 @@ extension SortedDictionary.SubSequence: Equatable where Value: Equatable {
   public static func ==(lhs: Self, rhs: Self) -> Bool {
     if lhs.count != rhs.count { return false }
     for (e1, e2) in zip(lhs, rhs) {
-      if e1.key != e2.key || e1.value != e2.value {
+      if e1 != e2 {
         return false
       }
     }
--- a/Sources/SortedCollections/SortedDictionary/SortedDictionary+SubSequence.swift
+++ b/Sources/SortedCollections/SortedDictionary/SortedDictionary+SubSequence.swift
@@ -300,7 +300,7 @@ extension SortedDictionary.SubSequence: Equatable where Value: Equatable {
   public static func ==(lhs: Self, rhs: Self) -> Bool {
     if lhs.count != rhs.count { return false }
     for (e1, e2) in zip(lhs, rhs) {
-      if e1 == e2 {
+      if e1.key != e2.key || e1.value != e2.value {
         return false
       }
     }
--- a/Sources/SortedCollections/SortedDictionary/SortedDictionary+Keys.swift
+++ b/Sources/SortedCollections/SortedDictionary/SortedDictionary+Keys.swift
@@ -239,7 +239,7 @@ extension SortedDictionary.Keys: Equatable {
   public static func ==(lhs: Self, rhs: Self) -> Bool {
     if lhs.count != rhs.count { return false }
     for (e1, e2) in zip(lhs, rhs) {
-      if e1 == e2 {
+      if e1 != e2 {
         return false
       }
     }
--- a/Sources/SortedCollections/SortedDictionary/SortedDictionary+Values.swift
+++ b/Sources/SortedCollections/SortedDictionary/SortedDictionary+Values.swift
@@ -258,7 +258,7 @@ extension SortedDictionary.Values: Equatable where Value: Equatable {
   public static func ==(lhs: Self, rhs: Self) -> Bool {
     if lhs.count != rhs.count { return false }
     for (e1, e2) in zip(lhs, rhs) {
-      if e1 == e2 {
+      if e1 != e2 {
         return false
       }
     }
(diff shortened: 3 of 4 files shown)


Sources/SortedCollections/SortedDictionary/SortedDictionary+SubSequence.swift
```diff
@@ -300,7 +300,7 @@ extension SortedDictionary.SubSequence: Equatable where Value: Equatable {
   public static func ==(lhs: Self, rhs: Self) -> Bool {
     if lhs.count != rhs.count { return false }
     for (e1, e2) in zip(lhs, rhs) {
-      if e1 == e2 {
+      if e1.key != e2.key || e1.value != e2.value {
```
xwu: Why not just `e1 != e2` here?

Hashim1999164: Good call. Switched it to `e1 != e2` so it matches Keys and Values.

lorentey (approved):
Thank you! 👍

PR #700, opened 2026-08-12
[Deque/HashTree] Remove references to ManagedBuffer.capacity and use the given minCapacity

Azoy:
When creating a `ManagedBuffer` object, the factory closure passes the newly created `ManagedBuffer` instance for the caller to make the header with. Referencing `ManagedBuffer.capacity` will end up making calls to `malloc_size` to return the allocation size, but the caller has already supplied it with a minimumCapacity argument. For embedded, it's critical that we don't needlessly link against system symbols that may or may not exist especially since the information we need is already there (and in fact the code was already guarding against it for OpenBSD).

### Checklist
- [X] I've read the [Contribution Guidelines](https://github.com/apple/swift-collections/blob/main/README.md#contributing-to-swift-collections)


lorentey (approved):
💯

PR #701, opened 2026-08-13
Name the documented arguments the way the declarations name them

karpovantonme:
25 doc comments in the shipping modules name an argument that the declaration below them does not have, so DocC has nothing to bind the description to and drops it. Nothing warns about this: Swift has no equivalent of `-Wdocumentation`, and DocC stays silent rather than complaining

Most are renames the comment did not follow, and the file itself usually holds the proof:

- `RigidDeque.prepend` documents `- maxCount` while its own `- Complexity: O(newItemCount ?? self.freeCapacity)` two lines below already uses the new name
- `RigidArray.append(copying:)` has 3 overloads. The oldest takes `newElements` and its comment is right; the 2 newer ones take `items` and carry the older one's wording
- `BitArray.index(after:)` writes the name in backticks, which DocC does not bind either, and the argument is `i`. Two functions below, `formIndex(after i:)` gets it right

3 entries name an argument that no longer exists at all rather than a renamed one, so they are removed instead: `stoppingOnFirstDuplicateValue` on `fill(untilFirstDuplicateIn:)` and `duplicates` on the two `_HashTable.create` overloads, where the behaviour now lives in the function name

where a description or a `- Complexity:` line referred to the old name too, it was brought along, so each block reads consistently

`SortedCollections` is left out on purpose: it sits behind the `UnstableSortedCollections` trait and has 17 more of these, which felt like a separate conversation. Happy to send that as its own PR if you want it

One I left alone: `UniqueArray.nextSpan(after:)` documents `maxCount`, but the whole block looks copied from the `nextSpan(maxCount:after:)` overload, and I could not tell whether you would rather drop the entry or the block

Comments only, no API change

btw these came out of a checker I wrote for exactly this, run over the tree and then read one by one. 6 of the things it flagged turned out to be my parser rather than your code, and those are fixed on my side rather than sent to you

### Checklist

### DIFF
--- a/Sources/BasicContainers/RigidArray/RigidArray+Append.swift
+++ b/Sources/BasicContainers/RigidArray/RigidArray+Append.swift
@@ -238,10 +238,10 @@ extension RigidArray {
   /// buffer, then this triggers a runtime error.
   ///
   /// - Parameters:
-  ///    - newElements: A fully initialized buffer whose contents to copy into
+  ///    - items: A fully initialized buffer whose contents to copy into
   ///        the array.
   ///
-  /// - Complexity: O(`newElements.count`)
+  /// - Complexity: O(`items.count`)
   @_alwaysEmitIntoClient
   public mutating func append(
     copying items: UnsafeMutableBufferPointer<Element>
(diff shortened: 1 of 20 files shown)


lorentey (approved):
Thank you for fixing all these! 👍

lorentey:

PR #704, opened 2026-08-14
Mark `_Word.allBits` as `@inlinable`

dnadoba:
Probably by accident `@inlinable` was left out here. Ideally we would eventually move to the official non underscore `@inline(always)` variant that implies `@inlinable` and would make this not look like it get's inlined but only within the module like it was before this patch. But that requires Swift 6.3+.

This has a surprisingly big effect on some algorithms. Over 10x for `BitArray.fill(with: true)` and `BitSet.fromUnion` given a Range:
PNG image

PNG image

Some other see "minor" improvements as well:

PNG image
PNG image
PNG image

### Checklist

### DIFF
--- a/Sources/InternalCollectionsUtilities/UnsafeBitSet/_UnsafeBitSet+_Word.swift
+++ b/Sources/InternalCollectionsUtilities/UnsafeBitSet/_UnsafeBitSet+_Word.swift
@@ -250,6 +250,7 @@ extension _UnsafeBitSet._Word {
     Self(0)
   }
 
+  @inlinable
   @inline(__always)
   package static var allBits: Self {
     Self(UInt.max)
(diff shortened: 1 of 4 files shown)


Sources/InternalCollectionsUtilities/UnsafeBitSet/_UnsafeBitSet+_Word.swift
```diff
@@ -250,6 +250,7 @@ extension _UnsafeBitSet._Word {
     Self(0)
   }
 
+  @inlinable
   @inline(__always)
   package static var allBits: Self {
     Self(UInt.max)
```
dnadoba, line 256: This is the only real change, the rest is just benchmarks.

lorentey (approved):
D'oh, good catch! 💯

PR #705, opened 2026-08-16
[SortedCollections] Name the documented arguments the way the declarations name them

karpovantonme:
#701 covers the shipping modules. This is the same thing in `SortedCollections`, which sits behind the `UnstableSortedCollections` trait and so is not in a default build

17 doc comments here name an argument that the declaration below them does not have, mostly renames the comment did not follow:

- `SortedDictionary.Keys` and `.Values` carry 8 blocks documenting `- Parameter i` while every one of those declarations takes `index`
- `_Node.UnsafeHandle` documents `newHandle` where the argument is `target`, and `newElement` where it is `newChild`
- `_Node.Splinter` documents `node`, the initializer takes `leftChild`
- `_BTree.Index` documents `index`, the initializer takes `offset`
- `_BTree+UnsafeCursor` documents `key` where `takeCursor(at:)` takes `index`

Where a `- Returns` line carried the old name too, it came along, so each block reads consistently. `index(after:)` and `index(before:)` in `Keys` and `Values` returned "the index immediately after `i`", and `takeCursor(at:)` returned "a cursor to the key or where the key should be inserted", which is the other overload's behaviour

The `i` in the neighbouring blocks is left alone on purpose: `index(after i:)`, `formIndex(_ i:)` and `index(_ i:, offsetBy:)` really are named that, and only these overloads were renamed without the comments following

One block is removed rather than renamed. `_BTree.swift` documents "a path to the key at absolute offset" with `- Parameter offset`, `- Returns` and `- Complexity`, then comes a blank line and then the doc comment for `startIndex(forKey:)`, so it binds to nothing

Comments only, no API change. Fine by me to close it if you would rather leave this module alone

### Checklist

### DIFF
--- a/Sources/SortedCollections/BTree/_BTree+UnsafeCursor.swift
+++ b/Sources/SortedCollections/BTree/_BTree+UnsafeCursor.swift
@@ -376,8 +376,8 @@ extension _BTree {
   /// This 'consumes' the tree, however it expects the callee to retain the root of the tree for the duration of
   /// the cursors lifetime.
   ///
-  /// - Parameter key: The key to search for
-  /// - Returns: A cursor to the key or where the key should be inserted.
+  /// - Parameter index: The index to take the cursor at
+  /// - Returns: A cursor to the element at that index.
   /// - Complexity: O(`log n`)
   @inlinable
   internal mutating func takeCursor(at index: Index) -> UnsafeCursor {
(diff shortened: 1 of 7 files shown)


lorentey (approved):
Thank you!

lorentey:

PR #718, opened 2026-09-03
[BitSet] Fix isEqualSet(to: Range<Int>) for empty ranges

dylanpulver:
`_toUInt()` returns nil whenever a bound is negative, and the guard maps that straight to `false`. That is right for a nonempty range — a bit set can never hold a negative value — but an empty range equals the empty set wherever its bounds sit, so `BitSet().isEqualSet(to: -5 ..< -5)` answered false.

On `main` that lets a bit set be both a subset and a superset of a range it calls unequal:

```swift
let s = BitSet()
s.isSubset(of: -5 ..< -5)    // true
s.isSuperset(of: -5 ..< -5)  // true
s.isEqualSet(to: -5 ..< -5)  // false
```

It also splits the generic overload against itself, since that one `_specialize`s `Range<Int>` back into this one: `isEqualSet(to: -5 ..< -5)` is false while `isEqualSet(to: Array(-5 ..< -5))` is true.

Five of the six `Range<Int>` predicates already handle it — `isSubset`/`isDisjoint` clamp, `isStrictSubset` composes those, `isSuperset`/`isStrictSuperset` check `other.isEmpty` before converting. This adds the same guard. `BitSet.Counted` forwards here, so it is covered too.

Clamping instead, to mirror `isSubset`, is tidier but wrong: it would make `BitSet([0, 1, 2]).isEqualSet(to: -5 ..< 3)` true.

New assertions extend `test_isEqual_to_integer_range` and fail on `main` with 3 failures. Full `swift test` is green (813 tests).

### Checklist

### DIFF
--- a/Sources/BitCollections/BitSet/BitSet+SetAlgebra isEqualSet.swift
+++ b/Sources/BitCollections/BitSet/BitSet+SetAlgebra isEqualSet.swift
@@ -41,6 +41,7 @@ extension BitSet {
   /// - Complexity: O(min(*max*, `other.upperBound`), where *max* is the largest
   ///    member of `self`.
   public func isEqualSet(to other: Range<Int>) -> Bool {
+    if other.isEmpty { return isEmpty }
     guard let other = other._toUInt() else { return false }
     return _read { $0.isEqualSet(to: other) }
   }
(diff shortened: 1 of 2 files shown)


lorentey (approved):
Good catch! 👍

PR #719, opened 2026-09-07
Add checkSetAlgebra, a law checker for set-like types

Joseph-Cursio:
`SetAPIChecker` verifies that a set-like type has the right *shape* — it's a protocol with no assertions, so conforming makes the compiler check that the API surface exists, including the generic-over-`Sequence` overloads `SetAlgebra` itself doesn't require. All five set types use it.

This adds the other half: that the operations behave the way the documentation says they do.

### Why a shared checker

That half is currently done by hand, once per type — 86 set-operation test functions across six files (25 in `OrderedSetTests`, 21 in `BitSetTests`, 13 each in `SortedSet Tests` and `BitSet.Counted Tests`, 12 in `TreeSet Tests`, 2 in `TreeDictionary.Keys Tests`; counted by test-function name, so a proxy for surface rather than an exact census).

Most already do the same thing the same way — model-based testing against `Set<Int>`. From `BitSetTests.test_union`:

```swift
withInterestingSets("a", maximum: 200) { a in      // a, b are Set<Int>
  withInterestingSets("b", maximum: 200) { b in
    let expected = a.union(b).sorted()
    expectEqualElements(BitSet(a).union(b), expected)
```

That's the same oracle a shared checker would use, written out once per operation per type.

### Design notes

**Constrained on `SetAPIChecker`, not `SetAlgebra`.** `OrderedSet` deliberately doesn't conform to `SetAlgebra` — documented at `OrderedSet.swift:100`, with `unordered` provided for the generic cases — so constraining on the stdlib protocol would exclude the type with the largest hand-written set-operation suite.

**The specification is validated before the type is judged**, following `checkEquatable`/`checkHashable`. A `Set<Element>` model is only as sound as `Element`'s `Hashable` conformance, so `_checkSetAlgebraModels` runs `_checkHashable` over the model elements and confirms each model survives a round trip through `Array` before any instance is checked; if it fails, the checker returns rather than reporting defects in the wrong artefact.

**Laws beyond agreement with the model**, chosen for cases where a type can be individually plausible and internally inconsistent: the `Sequence`-taking overloads checked separately from the `Self`-taking ones; commutativity, both absorption laws, and `symmetricDifference` against its definition in terms of the other three; the containment predicates against each other (`isSubset` against `isSuperset` reversed, `isDisjoint` against an empty intersection). Contents and count are asserted separately throughout, which matters for a type like `BitSet.Counted` that maintains a stored count.

Call sites for `BitSet`, `OrderedSet` and `TreeSet`; full suite passes.

### Verification

I checked that it can actually fail, by injecting defects into a stand-in conformance: an `isStrictSubset` that drops strictness and a `subtracting` that returns a stale count are both caught, and an unmodified control passes. The stale-count case is caught only because contents and count are asserted separately.

A `formUnion` that intersects passes every law here — the mutating `form*` laws that catch it are a follow-up PR, held until this one is settled since it depends on it.

### Questions

- Is the per-type approach deliberate for a reason I've missed? Happy to close this if so.
- `_checkSetAlgebraModels` is `package` rather than `private`, mirroring `_checkHashable`'s exposure, but nothing outside the file calls it — say the word and I'll make it private.
- The `filter` law uses `hashValue % 2` to get a non-trivial predicate. Both sides use the same values in the same process so it's stable, but if that reads as a seed dependency I can pass an explicit predicate instead.


Joseph-Cursio:
I was looking at ConformanceCheckers (and property-based testing in general) and noticed we have CheckSequence, CheckCollection, and CheckBidirectionalCollection, but didn't see anything for set algebra.

lorentey (approved):
Nicely done! Requiring `SetAPIChecker & Sequence` is a good compromise; `SetAlgebra` not being iterable was what blocked this previously.

PR #720, opened 2026-09-10
Validate the Container conformance of RigidDeque and UniqueDeque

Joseph-Cursio:
### What

Adds `test_validate_Container` to `RigidDequeTests` and `UniqueDequeTests`, so the last two `Container` conformers go through `checkContainer`.

### Why

Six types conform to `Container`. Four call `checkContainer` from their tests — `RigidArray`, `UniqueArray`, `RigidSet`, `UniqueSet`. The two deques do not, so nothing validates their conformance.

It isn't that they are untested. `RigidDequeTests.swift` and `UniqueDequeTests.swift` are substantial, but every assertion in them is written by hand, and none of it reaches the laws `checkContainer` covers: `nextSpan` boundaries stable across repeated iterations, a `BorrowingIterator` agreeing with `nextSpan` on both count and boundaries, `index(after:)` reaching `endIndex` in exactly `count` steps.

Each test enumerates every ring buffer layout of capacities 0, 1, 2, 3, 5 and 10 with `withEveryDeque(_:ofCapacities:)` — the same space `DequeTests.test_CollectionConformance` walks for `checkBidirectionalCollection`. That is the one place these differ from the array call sites, which pass a flat layout: for a deque the internal arrangement is the state that matters, and the wrapped layouts are where index arithmetic and span chunking would disagree if they were going to.

Both are behind the existing `#if compiler(>=6.4) && UnstableContainersPreview` gate, so they build and run exactly where the four existing call sites do. Two new tests, 26 lines, no source changes.

### Verified

Built with a `main` snapshot toolchain (2026-09-05) and run on `swiftlang/swift:nightly-main-jammy` with `--traits UnstableContainersPreview`: both new tests pass, alongside the four existing `test_validate_Container` cases.

Also checked that they can fail: replacing `expectedContents: data.contents` with `data.contents.dropFirst()` drives the checker into the contents comparison, so the tests are exercising `checkContainer` rather than passing vacuously. Worth noting what that produced, since it is a checker detail rather than anything about this change — the short expectation traps with `Fatal error: Index out of range` at `CheckContainer.swift:131` instead of reporting a validation failure. Fine for a deliberately broken oracle, but a caller who mis-sizes `expectedContents` gets a crash rather than a message.

### What to scrutinise

- **`expectEqual(tracker.instances, layout.count)`.** `LifetimeTracked` is a class, so the fixture array and the deque share instances: the count is `layout.count`, not the `2 * layout.count` in `RigidArrayTests.test_validate_Container`, where the expected-contents array is built separately. The idiom matches the rest of `RigidArrayTests`, which asserts `layout.count` after building a fixture from a layout.
- **Where the availability annotation sits.** `@available(SwiftStdlib 6.4, *)` is on the test method, following `RigidDequeTests.test_initCopyingBorrowingSequence` and `UniqueArrayTests.test_validate_Container`. `RigidArrayTests` instead uses `if #available` inside an ungated method. Happy to switch if the latter is preferred.
- **Cost.** 160 layouts per type, at most 10 elements each; both tests run in about 4ms.


Joseph-Cursio:
I noticed that we're testing:
* RigidArray and UniqueArray, 
* and RigidSet and UniqueSet,
* but not RigidDeque and UniqueDeque.

lorentey (approved):
Yep! 👍

Issue #429, opened 2024-10-25
[CI] Enable documentation check


shahmishal:
https://github.com/apple/swift-collections/actions/runs/11512888219/job/32048639180?pr=426
```
[405/405] Compiling Collections Collections.swift
Build of target: 'Collections' complete! (15.26s)
Generating documentation for 'Collections'...
Converting documentation...
warning: '--level' is deprecated. Use '--coverage-summary-level' instead.
warning: '--index' is deprecated. Use '--emit-lmdb-index' indead.
error: 'BitSet' doesn't exist at '/Collections'
  --> Collections.md:19:5-19:11
17 | ### Bit Collections
18 |
19 + - ``BitSet``
20 | - ``BitArray``
21 |

error: 'BitArray' doesn't exist at '/Collections'
  --> Collections.md:20:5-20:13
18 |
19 | - ``BitSet``
20 + - ``BitArray``
21 |
22 | ### Deque Module

error: 'Deque' doesn't exist at '/Collections'
  --> Collections.md:24:5-24:10
22 | ### Deque Module
23 |
24 + - ``Deque``
   |     ╰─suggestion: Replace 'Deque' with 'Deque-Module'
25 |
26 | ### Heap Module

error: 'Heap' doesn't exist at '/Collections'
  --> Collections.md:28:5-28:9
26 | ### Heap Module
27 |
28 + - ``Heap``
   |     ╰─suggestion: Replace 'Heap' with 'Heap-Module'
29 |
30 | ### Ordered Collections

error: 'OrderedSet' doesn't exist at '/Collections'
  --> Collections.md:32:5-32:15
30 | ### Ordered Collections
31 |
32 + - ``OrderedSet``
   |     ╰─suggestion: Replace 'OrderedSet' with 'Ordered-Collections'
33 | - ``OrderedDictionary``
34 |

error: 'OrderedDictionary' doesn't exist at '/Collections'
  --> Collections.md:33:5-33:22
31 |
32 | - ``OrderedSet``
33 + - ``OrderedDictionary``
   |     ╰─suggestion: Replace 'OrderedDictionary' with 'Ordered-Collections'
34 |
35 | ### Persistent Hashed Collections

error: 'TreeSet' doesn't exist at '/Collections'
  --> Collections.md:37:5-37:12
35 | ### Persistent Hashed Collections
36 |
37 + - ``TreeSet``
38 | - ``TreeDictionary``

error: 'TreeDictionary' doesn't exist at '/Collections'
[...]

PR #721, opened 2026-09-11
Unbacktick parameter names so DocC binds the parameter docs

hxperl:
### What's wrong

Six `- Parameter` entries write the parameter name inside backticks:

- `Sources/DequeModule/Deque/Deque+Collection.swift` — `index(after:)`, `formIndex(after:)`, `index(before:)`, `formIndex(before:)`, all as ``- Parameter `i`:``
- `Sources/BitCollections/BitSet/BitSet+BidirectionalCollection.swift` — `index(after:)`, `index(before:)`, both as ``- Parameter `index`:``

DocC binds parameter documentation by bare identifier, so a backticked name never
matches the parameter. The result is not a cosmetic one: **no Parameters section is
rendered at all**, and the entry falls through into the Discussion as a literal
`Parameter \`i\`: ...` bullet. Because the malformed bullet breaks the callout list,
the `- Complexity:` callout that follows gets absorbed into the same paragraph.

### Reproduction

On `main` at `774d6ba63114dd8302f2b509237748f54943ff2a`, before changing anything:

```
swift build --target DequeModule \
  -Xswiftc -emit-symbol-graph -Xswiftc -emit-symbol-graph-dir -Xswiftc /tmp/sg
xcrun docc convert --fallback-display-name DequeModule \
  --fallback-bundle-identifier org.swift.DequeModule \
  --additional-symbol-graph-dir /tmp/sg --output-path /tmp/out.doccarchive
```

`data/documentation/dequemodule/deque/index(after:).json` before:

```
sections = ['declarations', 'content', 'content']
   content: The next valid index immediately after `i`.
   content: Parameter `i` : A valid index of the deque. `i` must be less than `endIndex`. O(1)
```

There is no `parameters` section, and `O(1)` — the Complexity callout — has been
pulled into the same run of text. A correctly documented peer in the same module,
`Deque.subscript(_:)`, does produce one:

```
{"kind": "parameters", "parameters": [{"name": "bounds", "content": [...]}]}
```

`BitSet.index(after:)` and `index(before:)` show the same shape.

Worth noting: **DocC emits no warning for this.** `docc convert` exits 0 with an
empty diagnostic list, so neither a build nor the documentation check tracked in
 #429 would surface it. The rendered output is the only evidence.

### After

Same command on this branch, all six pages:

```
dequemodule/deque/index(after:).json      sections=['declarations','parameters','content','content']
   PARAM 'i':      A valid index of the deque. `i` must be less than `endIndex`.
dequemodule/deque/formindex(after:).json  sections=['declarations','parameters','content']
   PARAM 'i':      A valid index of the deque. `i` must be less than `endIndex`.
dequemodule/deque/index(before:).json     sections=['declarations','parameters','content','content']
   PARAM 'i':      A valid index of the deque. `i` must be greater than `startIndex`.
dequemodule/deque/formindex(before:).json sections=['declarations','parameters','content']
   PARAM 'i':      A valid index of the deque. `i` must be greater than `startIndex`.
bitcollections/bitset/index(after:).json  sections=['declarations','parameters','content','content']
   PARAM 'index':  A valid index of the bit set. `index` must be less than `endIndex`.
bitcollections/bitset/index(before:).json sections=['declarations','parameters','content','content']
   PARAM 'index':  A valid index of the bit set. `index` must be greater than `startIndex`.
```

Only the name in the `- Parameter` entry loses its backticks. Backticks inside the
description text are deliberately left alone — there they are correct.

### A judgement call

I put all six in one PR because they are one defect with one mechanical fix, and
splitting by module would make two PRs that have to be reviewed against the same
reasoning. If you would rather have Deque and BitCollections separately, say so
and I will split it.

### Tests

`swift test` on macOS 15 arm64, Swift 6.2: 18 XCTest suites, `Executed … with 0
failures (0 unexpected)` on every one, plus `Test run with 13 tests in 2 suites
passed`. For a comment-only change this is a regression check, not evidence for the
fix — the rendered DocC output above is the evidence.

Two honest caveats on that run. `swift test` exits 1 on this machine with `error:
Exited with unexpected signal code 11` interleaved into the `DequeTests` output; I
reran the identical command on a clean `main` with my changes stashed and got the
same exit code, the same signal, the same 18 suites and the same zero failures, so
it is pre-existing here and not something this change introduces. I have not
investigated it further and am not claiming it is or isn't a real bug. Separately,
the test bundle would not load at all until I copied `libswiftCompatibilitySpan.dylib`
from the Xcode toolchain next to it — a local toolchain issue, nothing to do with
the repo.

macOS only. CI will need to confirm Linux and Windows, though for a doc-comment
change I would not expect platform differences.

### Adjacent, deliberately not included

`Sources/ContainersPreview/Types/Shared.swift:62` has `- Returns true if this
instance was already uniquely referenced.` with no colon after `Returns`, so it is
also not parsed as a callout. Different file, different defect, left for a separate
change.

### Disclosure

This change was written with AI assistance. The reproduction, the DocC runs and the
test runs above are real output from my machine, not generated text.


lorentey (approved):
Nice, thanks!

PR #113, opened 2021-09-25
[Deque] `append(contentsOf:):` Use exponential capacity reservation

lorentey:
Depending on the supplied sequence, `Deque.append(contentsOf:)` currently reserves just enough capacity as it needs to perform the append. This isn't right -- it can result in O(n) reallocs instead of O(log(n)) when items are incrementally appended in constant-sized `append(contents:)` batches.

### Checklist
- [X] I've read the [Contribution Guidelines](/README.md#contributing-to-swift-collections)
- [X] My contributions are licensed under the [Swift license](/LICENSE.txt).
- [X] I've followed the coding style of the rest of the project.
- [ ] I've added tests covering all new code paths my change adds to the project (if appropriate).
- [ ] I've added benchmarks covering new functionality (if appropriate).
- [X] I've verified that my change does not break any existing tests or introduce unexplained benchmark regressions.
- [ ] I've updated the documentation if necessary.

PR #722, opened 2026-09-13
[DequeModule] Add tests for Deque's storage allocation behavior

### LINKED ISSUE #129: [Deque] Test storage allocation behavior
#113 fixed a performance issue with `Deque`'s storage allocation, but it did not add tests to cover this area.

Add some!

Joseph-Cursio:
Resolves #129.

#113 changed `append(contentsOf:)` to reserve capacity exponentially, but no tests cover `Deque`'s allocation behavior, and `RangeReplaceableCollectionTests.test_reserveCapacity` is still an empty `// FIXME: Implement` stub. This adds that coverage using the existing `_capacity` testing SPI. It changes tests only.

### What's covered

**`test_reserveCapacity` (implemented in place).** It runs over every layout of capacities `[0, 1, 2, 3, 5, 10]`, with minimum capacities 0 through 12, both unique and shared. It checks that a reservation never shrinks storage and grows it linearly to exactly `max(capacity, minimumCapacity)`. When unique storage already has enough room, it must stay in place. The contents must be preserved in every case.

**New `DequeAllocationTests`:**
- **Initializers with a known count allocate exactly that count.** This covers `init(minimumCapacity:)`, `init(_:)` from a collection and from a sequence with a precise `underestimatedCount`, and `init(repeating:count:)`.
- **Single-element insertions grow exponentially.** `append`, `prepend` and insertion in the middle each go through `0, 1, 2, 3, 5, 8, … 315` over 256 insertions. That's the same sequence `UniqueDequeTests.test_growth` expects.
- **Bulk insertions grow exponentially.** This is the #113 regression test. It covers 9 bulk operations: `append(contentsOf:)` and `prepend(contentsOf:)` from contiguous arrays, collections and sequences (including a sequence with no `underestimatedCount`), plus `insert(contentsOf:at:)` and `replaceSubrange` in the middle. Each runs 100 batches of 1, 2, 3 and 7 elements. Every reallocation must grow capacity by at least 1.5×, and the total must stay under 20 reallocations; linear growth would reallocate on nearly every call.
- **Reserved capacity is used without reallocating.** After `reserveCapacity(100)`, filling up to 100 elements with any of the above operations never changes capacity.
- **Inserting into shared storage.** The copy made for uniqueness keeps its capacity when there is room, and grows by the same factor as unique storage when full. The original is left untouched.

### Verification

`DequeTests` passes locally (124 tests) on Swift 6.4 / macOS.

To check the tests can actually fail, I injected two defects into `Sources/` and reverted them afterwards:
- **Linear growth.** Making `_growCapacity` always grow linearly fails the single-insertion, bulk-insertion and shared-storage tests.
- **Exponential reservation.** Having `reserveCapacity` pass `linearGrowth: false` fails `test_reserveCapacity`.

### Questions

- Some of this pins down current behavior exactly, such as the capacity sequence, exact-count initializers and linear `reserveCapacity`. That matches `UniqueDequeTests.test_growth`, but if any of it is meant to stay an implementation detail, I'm happy to loosen those checks to bounds.
- I deliberately left removals untested, since `remove(at:)` still has `// FIXME: Implement storage shrinking` and asserting that capacity doesn't shrink would lock that in.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016NTKVRe3aXMteHVrhbrTYE


lorentey (approved):
I like it! 

(I think `Deque` was using `malloc_size` at some point, which prevented tests from expecting concrete capacity numbers; but we are no longer depending on allocator behavior. 👍)

PR #730, opened 2026-09-22
Check the mutating form* twins against their non-mutating originals

Joseph-Cursio:
The follow-up promised in #719, now that it has landed.

#719 checks the non-mutating set operations against a `Set<Element>` model. It does not check the mutating ones at all — `formUnion`, `formIntersection`, `subtract` and `formSymmetricDifference` appear in `SetAPIChecker`, which verifies only that they *exist*. So **a `formUnion` that simply intersects passes every law #719 ships.**

This adds the missing half: each mutating operation is checked against its own non-mutating twin.

### What it checks

The four pairs — `formUnion`/`union`, `formIntersection`/`intersection`, `subtract`/`subtracting`, `formSymmetricDifference`/`symmetricDifference` — in both the `Self`-taking and `Sequence`-taking overloads, so eight mutations per pair of inputs.

Each runs under `withHiddenCopies` at both sharing states. A mutating operation on a shared value is where copy-on-write bugs live, and afterwards the original is checked to confirm it was not mutated through the copy.

Contents and count are asserted separately, matching the rest of the checker. That is what catches a type like `BitSet.Counted` updating its elements correctly and its stored count wrongly.

### Why this is the pair worth having

The two implementations should agree by construction, so a divergence is a real bug rather than a debatable one — unlike a law where a type might reasonably have its own opinion. It is also the only thing here that constrains the mutating half at all.

### Design notes

**No new call sites.** This extends `_checkSetAlgebraBinaryLaws`, so the three existing `checkSetAlgebra` callers — `BitSetTests`, `OrderedSetTests`, `TreeSet Tests` — pick it up with no change. The diff is one file.

**Cost.** Eight mutations at two sharing states per input pair. The three `test_SetAlgebraConformance` cases each run in under 10ms.

### Verification

Verified by injecting three defects into a stand-in conformance: a `formUnion` that intersects, an `isStrictSubset` that drops strictness, and a `subtracting` that returns a stale count. All three are caught, and an unmodified control passes. The stale-count case is caught only because contents and count are asserted separately.

I also checked that the new block is not passing vacuously, by expecting `formUnion` to equal the intersection: it fails on both the contents and the count assertion, at both sharing states.

Rebased onto current `main` and re-run: full suite green, 821 tests, 0 failures.

### Questions

- **Both sharing states, or just shared?** Running at `isShared == true` alone would halve the work and is where the interesting bugs are. I kept both because the unshared path is the one most types optimise separately, but I don't feel strongly.
- **Is the "original was not mutated" assertion redundant?** `withHiddenCopies` already exists to catch writes through a shared buffer, so this may be belt-and-braces. Happy to drop it if it duplicates what the helper guarantees.
- **What's the preferred way to reach the `Sequence`-taking overloads?** I pass `Array(ym)`, since `y` is `Self` and overload resolution would send it to the `Self`-taking overload instead. An `Array` of the model's elements was the most obvious way to get the same values in under a different static type, but if there's an idiom for this elsewhere in the checkers I'd rather match it.


lorentey (approved):
Looks good! Thank you for doing these 👍

Joseph-Cursio:
I'm learning a lot about property-based testing, and how Apple does it. The approach to not use generators is certainly different than what I've seen elsewhere.

